PR #56, opened 2026-09-17
ledger: record run 0007 and where a missing guard gets fixed

abdullah-life:
Records `run/0007`. `item/content-actions` complete via [#55](https://github.com/turbophp/doccum/pull/55); `item/trash-directory` now active.

**9 of 35 v1 items done.**

## `decision/0009` — a pre-existing action missing a guard is fixed inside the item that names the guard

`MoveDirectory` had no name collision check at all. Moving a directory into a parent already holding that name silently succeeded, leaving two live siblings sharing a name — the state spec §4 says cannot exist. `CreateDirectory` checked; the move path never did.

The rescoped `content-actions` issue said `MoveDirectory` "already exists" and needed only refusal-test coverage. That was wrong, and I wrote it. The subagent implementing the item noticed its brief and the issue disagreed and **flagged the contradiction rather than silently resolving it either way**, which is the only reason it surfaced.

The fix belongs in that item, not a new one: the item's own `doneWhen` says a rename *or move* into a taken name throws the existing Duplicate exceptions, and `MoveDirectory` is one of the two move paths. Filing it separately would have satisfied that sentence for `MoveFile` while the directory equivalent stayed broken.

Recorded as a rule: when an item names a guarantee, a pre-existing action that silently lacks it is inside that item's scope, and the **issue text gets corrected** rather than the scope narrowed to match what was already written.

## Also in the run

`FilePolicy::move()` gates on directory access with no Spatie permission, which looks like a two-layer violation and isn't — there is no file-*editing* capability to check against, and the pre-existing `update()`, `view()` and `download()` are access-only for the same reason. Recorded in the run description so the next reader doesn't re-litigate it.

## Verification

`php .github/scripts/validate-ledger.php` → **Ledger is sound.**

Written in a separate git worktree, since the `trash-directory` agent is working in the main checkout and a branch switch would have pulled files out from under it.

---


abdullah-life:
## `image` failed here, and it is not this diff

**Failing check:** `image` — [run 35212200191](https://github.com/turbophp/doccum/actions/runs/35212200191/job/105172111848)

```
extraction reached a terminal non-success status (failed) for smoke-35212200191-1.txt
container-smoke (setup) failed: extraction status "failed" for smoke-...txt
```

`App\Jobs\ExtractText` failed inside the container. The container itself built, booted, ran every migration and answered the installer — supervisor shows `minio`, `frankenphp`, `worker` and `worker-ingest` all in `RUNNING` state.

**Why this is not this PR's:** the diff is two JSON files under `docs/ledger/`, which nothing in the container reads. The same application code is on `main` at `3702b8f` (the #55 merge), and [run 35211857569](https://github.com/turbophp/doccum/actions/runs/35211857569) — including its own `image` job — passed at 10:42, five minutes before this failure.

Re-running the failed job once to confirm, which is the one re-run this warrants.

## This is still worth a real answer

`ExtractText` failing on a plain `.txt` upload is not a plausible "infrastructure" outcome — the file needs no `tesseract`, no `pdftotext`, nothing external. So either it is a genuine intermittent bug in the ingest path, or the smoke test is racing the object store.

If the re-run passes, I will file that as its own issue rather than let it disappear. An intermittent extraction failure is exactly the kind of thing a self-hoster would hit and be unable to diagnose, and the whole reason the smoke test exists is that "a green suite is necessary and not sufficient". The container job has now earned its keep twice.

---

Issue #65, opened 2026-09-17
Six foreign key columns have no index outside MySQL


abdullah-life:
**Ledger id:** `item/fk-index-coverage`
**Size:** S
**Depends on:** `item/schema-query-audit`

Found by `item/schema-query-audit` ([#60](https://github.com/turbophp/doccum/issues/60)) — the first thing that audit turned up, which is the point of it existing.

## The gap

Six foreign key columns have **no index whose leading column is that column**:

| column | why |
|---|---|
| `directories.created_by` | `foreignId(…)->constrained('users')`, no index of its own |
| `files.created_by` | same pattern |
| `file_versions.uploaded_by` | `file_id` is covered by the `(file_id, version_number)` unique index; `uploaded_by` is not |
| `search_documents.owner_id` | `directory_id` is indexed explicitly, `owner_id` is not |
| `settings.updated_by` | no index at all |
| `role_has_permissions.role_id` | spatie's composite primary key `(permission_id, role_id)` leads with `permission_id`, so `role_id` is never a leading column |

**MySQL hides this.** InnoDB auto-creates an index when a foreign key requires one, so all six are covered there and invisible. SQLite and PostgreSQL leave the columns entirely unindexed — verified empirically against both (`PRAGMA index_list`, `pg_indexes`) while building the audit, not inferred from reading the migrations.

## Why it matters

Unindexed FK columns make deletes and lookups against them scan. `settings.updated_by` and `search_documents.owner_id` both carry `nullOnDelete()`, so deleting a user rewrites those tables with no index to find the rows by.

It is also the mirror image of `decision/0008`: that was an index silently *carrying* a constraint and being dropped; this is a constraint with no index to carry it on two of three drivers. Same blind spot, opposite direction — and again only one engine tells you.

## Done when

1. Each of the six has an index leading with the FK column, or a recorded reason it does not need one. `role_has_permissions.role_id` is a third-party migration, so it likely needs an additive migration of our own rather than an edit — `CLAUDE.md` forbids editing a framework migration, and the same reasoning applies to a package's.
2. The corresponding entries are **removed** from `SCHEMA_AUDIT_FK_INDEX_EXCLUSIONS` in `tests/Feature/SchemaPortabilityTest.php`, so the audit asserts the rule rather than recording the exception.
3. Any new index is created before any index it supersedes is dropped, per `decision/0008`.

Clause 2 is the real deliverable: the audit is only as strong as its exclusion list is short.

---
[...]

Issue #60, opened 2026-09-17
Audit every semantic doccum inherits from a driver default instead of stating


abdullah-life:
**Ledger id:** `item/schema-query-audit`
**Size:** M
**Depends on:** nothing

## The pattern, stated once

Six defects so far share one root cause, and it is not "MySQL is odd". It is that **the code inherited a semantic from a driver or framework default instead of stating it**:

| defect | inherited default that decided behaviour |
|---|---|
| [#37](https://github.com/turbophp/doccum/issues/37)/[#38](https://github.com/turbophp/doccum/issues/38) — MySQL uninstallable | InnoDB's 3072-byte index key limit |
| [#2](https://github.com/turbophp/doccum/pull/2) — PostgreSQL search silently empty | PostgreSQL's case-sensitive `LIKE` |
| [#46](https://github.com/turbophp/doccum/issues/46) — filename uniqueness | MySQL's `utf8mb4_unicode_ci` folding |
| `decision/0008` — errno 1553 | InnoDB requiring an index on a FK column, satisfied by an index nobody knew was carrying it |
| [#39](https://github.com/turbophp/doccum/issues/39) — latent, non-ASCII only | PostgreSQL's ~2704-byte btree tuple limit |
| [#49](https://github.com/turbophp/doccum/issues/49) — trashed subtree still searchable | Eloquent's `SoftDeletes` global scope accidentally acting as the reach filter |
| [#59](https://github.com/turbophp/doccum/issues/59) — email identity | the column collation again |

`decision/0008` recorded the index-ordering rule. That is a **special case**. The general rule belongs in the ledger and the audit belongs in a test.

## Done when

1. **`SchemaPortabilityTest` runs on all six legs.** Using `Schema::getIndexes()`/`getColumns()` it computes each index's worst-case key width (declared varchar length × 4 for utf8mb4) and asserts ≤ 3072 on `mysql` and ≤ 2704 on `pgsql`, or that the index appears on an explicit list of driver-conditional expressions. It also inserts a string of each indexed varchar column's declared maximum length built from 4-byte characters and reads it back — **this must be shown failing on `pgsql` for `properties.value_string` first**, which closes [#39](https://github.com/turbophp/doccum/issues/39) with the proof that issue asks for.
2. **Every foreign-key column has an index leading with that column**, asserted via `Schema::getForeignKeys()` + `getIndexes()` on all legs. This generalises `decision/0008` so no future index swap can silently drop the one carrying a constraint.
[...]

PR #67, opened 2026-09-17
ledger: record run 0009 and the item the audit found on its first run

abdullah-life:
Records `run/0009`. `item/schema-query-audit` complete via [#66](https://github.com/turbophp/doccum/pull/66) — **11 of 42 v1 items**.

## What the audit buys

The class of defect behind seven bugs now fails at CI rather than in an install. From here on, on every matrix leg:

- a new `DB::statement`, `whereRaw`, `LIKE` or `LOWER()` under `app/` or `database/` **fails** unless registered with the driver semantics it relies on;
- a new index over InnoDB's 3072 or PostgreSQL's 2704 bytes fails on the engine that enforces it;
- a new foreign key without a leading index fails;
- a new **driver** added to the matrix without a registered key-length ceiling fails, rather than silently opting out.

## It found something immediately, then found something in itself

`item/fk-index-coverage` ([#65](https://github.com/turbophp/doccum/issues/65)) is added at order 12: six foreign key columns with no index leading with that column outside MySQL, where InnoDB auto-creates one and hides the gap. Filed rather than fixed inline — this item asserts, it does not fix.

Then the audit's **own** first CI run failed: it asserted index width on SQLite, which has no key length ceiling at all, tripping over SQLite's implicit autoindex for the framework `cache` table. The skip was written to *name* the driver rather than rely on its absence from the limits table, so a driver added later fails instead of quietly escaping a check it was meant to be under.

Both are the same argument for the item, made twice in one run.

## The process finding

**Three items have now had a `doneWhen` that did not survive contact with the code**: `content-actions` split three ways, `trash-directory` turned out to be a defect rather than missing work, and this one demanded a proof that could only be given in a different item.

This one was caught *before* delegation rather than after — the clause requiring a multi-byte round-trip "shown failing on PostgreSQL first" would have meant merging a red leg, since fixing that failure is `pgsql-value-string-index`, a separate item depending on this one. The proof moved there and [the reasoning went on the issue](https://github.com/turbophp/doccum/issues/60#issuecomment-5713559581).

Every remaining `doneWhen` was written at `run/0000` from the spec, before any of the code existed. Checking each clause against `main` before delegating is the most valuable step in an iteration, and it is now treated as one.

## Verification

`php .github/scripts/validate-ledger.php` → **Ledger is sound.**

---


abdullah-life:
## Both PostgreSQL legs are red, and it is not this diff

**Failing:** `test (php 8.4, pgsql)` and `test (php 8.5, pgsql)` — `TrashDirectoryTest > it stamps every cascaded row with the same deleted_at as the root`, *"Failed asserting that two DateTime objects are equal."*

This PR changes two JSON files under `docs/ledger/`. It cannot affect `TrashDirectory`.

**The cause is a race I introduced in [#63](https://github.com/turbophp/doccum/pull/63).** That assertion was correct while the cascade pinned an identical `deleted_at` across every row — and I removed the pinning in the same PR, when `trashed_batch` replaced it, precisely because a timestamp cannot tell a cascade from a file trashed in the same second.

Since then each row takes its own `deleted_at` at `delete()` time, so comparing them passes **only while the whole cascade lands inside one tick of the column's precision**. Green by luck on fast SQLite and MySQL runs; PostgreSQL, storing microseconds, finally crossed a boundary. Three green runs hid it.

**Fix: [#68](https://github.com/turbophp/doccum/pull/68)**, already open — both tests now assert `trashed_batch`, which is what actually carries "these rows went together" at any timing.

**Not ported into this PR deliberately.** `docs/LOOP.md` keeps ledger commits on a `ledger/` branch and out of feature PRs; carrying a test change here to go green sooner would invert that for no gain, since #68 is minutes from merging. This branch will be updated from `main` once it lands, and re-run without a byte of the ledger changing.

No re-run spent: a re-run cannot fix this, and the failure is real rather than flaky — it is a race that will recur until #68 lands.

---

Issue #59, opened 2026-09-17
Email uniqueness means different things on different databases — the same bug as #46, one table over


abdullah-life:
**Ledger id:** `item/email-folding`
**Size:** S
**Depends on:** nothing

## Verified against `main`

`ProfileValidationRules::emailRules()` (`app/Concerns/ProfileValidationRules.php:39`) ends with:

```php
$userId === null
    ? Rule::unique(User::class)
    : Rule::unique(User::class)->ignore($userId),
```

No normalisation. Uniqueness is therefore decided by the column collation — the identical mistake [#46](https://github.com/turbophp/doccum/issues/46) closed for file and directory names, one table over and still open.

| | register `Alice@example.com` when `alice@example.com` exists |
|---|---|
| MySQL (`utf8mb4_unicode_ci`) | rejected — **one account** |
| SQLite / PostgreSQL | accepted — **two accounts**, and a differently-cased login fails |

`username` is safe, but only by accident: its regex forces lowercase ASCII, so no folding question arises.

## Why this is worse than the filename version

Two accounts differing only in case is an identity problem, not a tidiness problem. On SQLite and PostgreSQL a user who registered as `alice@` and later types `Alice@` is told their credentials are wrong, with no way to discover why. On MySQL the same pair is one account — so the *same instance data* behaves differently depending on which database the operator chose at install time.

## Done when

1. Email is folded explicitly in the application, not left to the collation. **Recommend lowercase-on-write**, ratifying what MySQL already does — the same call [#46](https://github.com/turbophp/doccum/issues/46) made for names, and consistent with how essentially every mail provider treats the local part in practice.
2. Login resolves the user through the same folding rule, so a differently-cased login succeeds on every driver.
3. A test asserting the chosen semantics runs on **all six matrix legs** and is shown failing on at least one driver before the fix — per the rule established in `decision/0006`.
4. The choice is recorded, since it is a behaviour change for whichever driver loses.

Note the fix is *not* simply reusing `Support\NameKey`: that is case-insensitive **accent-sensitive** for filenames, and email needs its own stated rule rather than a borrowed one. State it, do not inherit it.

Found during the periodic architecture review at 9/35 v1 items — the fourth instance of the same root cause, which is why `item/schema-query-audit` is being filed alongside it.

---
Tracked by the development loop — see `docs/LOOP.md`. Status of record is `docs/ledger/ledger.jsonld`.

abdullah-life:
[...]

PR #77, opened 2026-09-17
[EXPERIMENT — do not merge] email folding tests without the fix

abdullah-life:
**Throwaway. Not for merge — opened only to make CI run, which it does not do for branches without a PR.**

`item/email-folding` (#59) has a `doneWhen` clause requiring the test be *"shown failing on at
least one driver first"*. `composer install` cannot complete in the dev environment, so nobody
can demonstrate that locally. This branch demonstrates it in CI instead.

It carries the tests from `claude/email-folding` and the pure `App\Support\EmailKey` helper,
but **not** the behaviour change: no folding on write in `CreateNewUser`/`FirstRun`/`Profile`,
no `Fortify::authenticateUsing()` wiring, no `AuthenticateUser` action.

## The prediction being tested

| Leg | Expected | Why |
|---|---|---|
| sqlite 8.4 / 8.5 | **RED** | byte comparison on `=`; no collation folding |
| pgsql 8.4 / 8.5 | **RED** | same |
| mysql 8.4 / 8.5 | **GREEN** | `utf8mb4_unicode_ci` already folds case |

If MySQL also fails, or SQLite passes, the analysis behind the fix is wrong and the fix gets
rethought rather than merged.

`tests/Feature/SchemaPortabilityTest.php` may also fail here for an unrelated reason — its
folding-rule inventory is updated to the post-fix state. That is noise; the signal is the four
named behavioural tests.

Closed once it has reported.

---


abdullah-life:
## Result — the fix is justified, my prediction table was not

All six legs went red, where I predicted MySQL would stay green. The reason is that these tests
fall into **two classes**, and I conflated them:

| Leg | Failures | Which |
|---|---|---|
| SQLite / PostgreSQL | **6** | storage-folding **and** login/uniqueness |
| MySQL | **3** | storage-folding only |

**MySQL failures** — `a registered email is stored folded to lowercase`,
`it folds the first admin's email to lowercase…`, `profile email is folded to lowercase when updated`.
These assert the *stored value*. `utf8mb4_unicode_ci` makes comparison case-insensitive but does
not change the bytes on disk, so they fail on every driver without the fix.

**SQLite/PostgreSQL adds** — `registering an email that only differs by case…`,
`it lets the first admin log back in with different case`, `updating email to a case variant…`,
`user can log in after changing their email to mixed case`. These are the cross-driver ones, and
MySQL's collation masks every one of them.

That is exactly the split the implementing agent described, and its four named tests do pass on
MySQL as it claimed. `it lets the first admin log back in with different case` failing on SQLite
is the installer lockout, demonstrated pre-fix.

Purpose served. Closing unmerged; the fix is #80.

---

PR #78, opened 2026-09-17
[EXPERIMENT — do not merge] mutation: remove @fluxScripts to probe the smoke

abdullah-life:
**Throwaway. Not for merge — opened only to make CI run.**

CLAUDE.md requires every guard be mutation-checked. The container smoke's assertions never
have been, including the ones I added in #72.

#72 claimed that asserting the account menu hidden-before-click and visible-after proves
`@fluxScripts` booted in the image. That claim is untested and may well be **wrong**: if Flux
renders the menu as a native popover driven by a `popovertarget` attribute, the browser hides
it and opens it on click with no JavaScript at all — in which case both halves pass on a
container shipping no scripts, and the check proves nothing.

This removes `@fluxScripts` from `resources/views/layouts/app/topbar.blade.php`.
`checkTopbar()` runs **before** the upload step in the `setup` phase, so the failure point is
the answer:

- **fails inside `checkTopbar`** → the dropdown assertion is load-bearing, #72 stands
- **reaches the upload step** → it is not, and both #72's comment and the CLAUDE.md sentence
  built on it ("assert the hidden state before the interaction") are wrong and get corrected

Either way the `image` job should go red — removing `@fluxScripts` also breaks Livewire's
client, so the upload will fail if the topbar check doesn't. **Where** it goes red is the
entire point; a green `image` job here would mean the smoke is far weaker than anyone thinks.

Closed once it has reported.

---


abdullah-life:
## Result — the dropdown check is load-bearing, but not for the reason #72 gave

The `image` job failed **inside `checkTopbar`**, on the click:

```
[setup] version pill renders the configured value: v0.1.0
[setup] topbar shows Home, Files and Settings for the administrator
```

`account menu opens on click` never printed, and the run never reached the upload step.

**So "menu opens on click" genuinely proves the scripts loaded and booted in the image.**

But note what did *not* happen: the hidden-before-click check passed. In an image with **no
JavaScript at all**, the menu was still hidden. So that half proves nothing on its own — the
opposite of what #72 claimed and of what I then wrote into CLAUDE.md as a rule for the remaining
14 UI items.

The version pill check also passed here, as it must: it is server-rendered and compares
`config('doccum.version')` against itself.

Corrections in #81. Closing unmerged.

---

Issue #16, opened 2026-09-16
Home dashboard: recents, storage by period, quick search


abdullah-life:
**Ledger id:** `item/home-dashboard`
**Spec:** [10-user-facing-surface](https://github.com/turbophp/doccum/blob/main/docs/superpowers/specs/2026-09-15-doccum-design.md#10-user-facing-surface)
**Size:** S
**Depends on:** `item/topbar-shell`

### Done when
Recents are filtered by DirectoryAccess so another user's file is absent; the page renders with zero data.

---
Tracked by the development loop — see `docs/LOOP.md`. Status of record is `docs/ledger/ledger.jsonld`.

abdullah-life:
Shipped in turbophp/doccum#160 (`98d7a7e`), the first item on run/0019. Accounting per `doneWhen` clause — including the parts of spec §10's Home that are deliberately **not** built.

### "Recents are filtered by DirectoryAccess so another user's file is absent"

`Home\Index` resolves `viewableDirectoryIds()` through the seam and narrows with `whereIn` **in the query** — never a Blade `@if`, never a collection filter after the fact. Asserted against `viewData('recentFiles')` directly, not just against page text.

Trashed files fall out of `File`'s own `SoftDeletes` scope rather than an extra condition — the worker confirmed that by reading the model rather than assuming it, and it is tested both ways (a trashed file excluded, a live file in the same directory still listed).

`.github/mutations.json` gained `home-recent-files-query-scoped-to-reach`, so `guards` re-proves the `whereIn` every run: without it, Home would list the most recently updated files across the whole instance.

### "the page renders with zero data"

Tested twice — component-level and over HTTP. A brand-new user with no reachable directory gets the empty state, not a crash and not a blank section.

### "Exactly one link on /files has the accessible name 'Home'"

See #140. Proven mechanically rather than demonstrated once; the `doneWhen` asked for it "shown failing against main first", and neither the worker nor I could run Pest without `vendor/`, so it became a `mutations.json` entry instead — which `guards` has already run in both directions.

### Container smoke, mutation-proven

`checkHomeDashboardShowsRecentUpload()` uploads a file and requires it to appear under Recent files — real data through the query into the rendered page, not a resting state. Mutation #162 left the component untouched and stopped the view rendering rows; `image` went red inside that check with every prior check green.

[...]

PR #81, opened 2026-09-17
docs: correct what the topbar smoke assertions actually prove

abdullah-life:
#72 made three claims about its own assertions. **Two were wrong**, none had been tested, and
CLAUDE.md then codified the wrong one as a standing rule for the remaining 14 UI items.

## The experiment

`@fluxScripts` removed from the layout, smoke run against that image (#78):

https://github.com/turbophp/doccum/actions/runs/35223275581

```
[setup] version pill renders the configured value: v0.1.0
[setup] topbar shows Home, Files and Settings for the administrator
```

It failed **inside `checkTopbar`**, on the click — `account menu opens on click` never printed.

## What that settles

| Claim in #72 | Verdict |
|---|---|
| "menu opens on click" proves the scripts booted | ✅ **Confirmed** — load-bearing |
| "hidden before the click" is what makes it load-bearing | ❌ **Wrong** — with *no JS at all* the menu was still hidden, so this half passes on a container shipping no scripts |
| version pill guards against a build-time config cache | ❌ **Wrong** — `tinker` and the Blade both read `config('doccum.version')`: same process, same source, and the Dockerfile runs no `config:cache` |

The version pill check still proves something narrow and real — the pill renders the configured
value rather than a literal baked into the view. It becomes a genuine cross-source comparison
once `item/version-from-tag` (36) supplies the image's `org.opencontainers.image.version` label.

## Changes

- `checkTopbar()`'s docblock now states what each assertion is worth **as measured**, and links
  the mutation run.
- The hidden-before-click check keeps its place as a cheap sanity check, explicitly labelled as
  proving nothing on its own.
- The CLAUDE.md rule is rewritten around the lesson rather than the mistake: *a smoke assertion
  is worth only what a mutation says it is*; break the feature deliberately, run the smoke, record
  the run; prefer an assertion that requires the feature to **do** something over one observing a
  resting state; never cite an unmutated assertion as evidence.

No behaviour change — comments, log wording and the rule. The mutation branch is closed, not merged.

---


abdullah-life:
## `image` is red, and it is not this PR's

This PR changes comments in `container-smoke.mjs` and one rule in `CLAUDE.md`. No behaviour.

The topbar assertions this PR documents all **passed**:

```
[setup] version pill renders the configured value: v0.1.0
[setup] topbar shows Home, Files and Settings for the administrator
[setup] account menu opens on click -- scripts booted in the image
[setup] upload accepted, file listed in the directory
```

It then failed further on, in text extraction:

```
##[error]extraction reached a terminal non-success status (failed)
2026-09-17 12:55:55 App\Jobs\ExtractText .... FAIL
```

The container logs show the workers spawned at `12:55:28` and `ExtractText` failing at
`12:55:55`, with MinIO only just having entered `RUNNING`. That is a transient read against
storage that is not serving yet, being treated as a **terminal** failure — which is exactly
`item/extract-text-flake` (order 16, #16's scope), still unstarted. The same `image` job passed
on #79 minutes earlier against the same base, so it is intermittent rather than a regression.

No fix exists to port yet, so there is nothing for me to carry into this PR. I have started
item 16 rather than leaving this to recur; it is now demonstrably costing CI cycles that nobody
can reproduce locally, since `composer install` cannot authenticate in the dev environment.

Re-running the job once, per the one permitted confirmation re-run.

---

PR #88, opened 2026-09-17
[EXPERIMENT — do not merge] mutation: remove the failed-response mail guard

abdullah-life:
**Throwaway. Not for merge — opened only so CI runs.**

CLAUDE.md now requires that a smoke assertion be **shown** to fail when the feature is broken, not
argued to. That rule exists because the previous round's reasoning about which half of an assertion
carried the weight was proven exactly backwards by experiment.

This branches off `claude/user-reset-password-command` and removes the `MailDeliverability` guard
from the **failed** password-reset-link response only, leaving the success path intact.

## What should happen

With the guard gone, the success path still renders "Mail is not configured" while the failure path
falls through to Fortify's stock validation error. The two responses **diverge** — which is exactly
the account-existence oracle the feature exists to close.

So:

- the PHP test comparing the two responses should fail on all six legs
- the container smoke's forgot-password check should fail looking for a status node that is no
  longer rendered for the unknown address

**If either stays green, that assertion proves nothing and gets rewritten before the item lands.**

Closed once it has reported.

---


abdullah-life:
## Result — the assertions are load-bearing

Three tests failed on every leg, all in `ForgotPasswordMailDeliverabilityTest`, and the failure text
is the account-existence oracle itself:

```
Session has unexpected errors:
  "We can't find a user with that email address."
```

plus the JSON comparison: `Expected response status code [200] but received 422`.

With the guard removed, the unknown address gets Laravel's stock "we can't find a user" while the
real address still gets "Mail is not configured" — the two responses diverge, which is exactly the
oracle #89 exists to close, and the assertions catch it. `592 passed, 3 failed`.

So those assertions test what they claim. Proved by experiment rather than argued.

**One artifact worth noting**: `lint` failed first, on a Pint style issue my own edit introduced when
it left the guard's braces behind. That is a property of how I made the mutation, not a signal about
the feature — the Pint bot reformatted it and the test legs then gave the real answer. A mutation
that trips a formatter before reaching the tests tells you nothing; worth making the cut cleanly next
time.

Closing unmerged. Next experiment reuses this branch for the reset-link token assertion.

---

abdullah-life:
## Result — mutation 2, and the full record

The reset token was corrupted so the command still exits 0 and still prints a correctly-shaped link.

**The container smoke failed at exactly the predicted point:**

```
[setup] checking doccum:user:reset-password prints a working link with no mailer configured
[setup] doccum:user:reset-password printed a link for smoke-reset-…@example.test
##[error]container-smoke (setup) failed: page.waitForURL: Timeout 10000ms exceeded.
```

It printed the link, then failed *driving* it — which is the distinction that matters. A check on the
exit code or the URL's shape would have stayed green on a link that cannot reset anything.

**All six test legs also failed**, on `UserResetPasswordTest`: *"This password reset token is
invalid."* I had predicted only `image` would go red, assuming that PHP test merely checked the
printed string. It does not — it drives the printed link through the real reset flow. The test was
stronger than I credited.

### Full record for #89

| Assertion | Mutation | Result |
|---|---|---|
| Forgot-password PHP tests catch the oracle | guard removed | **red** — failure text *was* the leak |
| Container smoke catches the oracle | guard removed | **red** (`image`) |
| Reset-link PHP test drives the real flow | token corrupted | **red** — all six legs |
| Container smoke drives the reset flow | token corrupted | **red** — timed out on the reset form |

Four for four, proven rather than argued.

### Two notes against my own method

1. Mutation 1 tripped **Pint** before reaching the tests, on formatting my own edit introduced. A
   mutation that fails a formatter first tells you nothing; I only got the answer because the Pint bot
   reformatted and CI re-ran.
2. Mutation 2's run also carried a pre-existing failure in `ForgotPasswordMailDeliverabilityTest`,
   because this branch predates the session-ordering fix on #89. Noise in the signal — a mutation
   branch should be cut from the head it is testing, not from an earlier one.

Closing unmerged.

---

Issue #92, opened 2026-09-17
SQLite lock contention is the real cause of the extraction flake, and it is a driver default left unstated


abdullah-life:
## The cause, printed on `main`

`item/extract-text-flake` (#57) was closed saying the cause was unknown. **It was not.** The
diagnostic added by that very item printed it on `main` twice within 25 minutes of the item closing
— runs [35227625151](https://github.com/turbophp/doccum/actions/runs/35227625151) and
[35229499626](https://github.com/turbophp/doccum/actions/runs/35229499626):

```
file_texts.error: SQLSTATE[HY000]: General error: 5 database is locked
  (Connection: sqlite, Database: /var/www/html/database/database.sqlite,
   SQL: update "jobs" set "reserved_at" = 1789651959, "attempts" = 1 where "id" = 4)
```

Not an object-store race. Not MinIO readiness. **SQLite write contention on the `jobs` table.**

## Why it happens, and why it is `decision/0010` again

The single container runs FrankenPHP, two `queue:work` processes and `schedule:work` against one
SQLite file, with **queue, cache and session all on the database driver**:

- `config/queue.php:16` — `database`
- `config/cache.php:18` — `database`
- `config/session.php:21` — `database`

And `config/database.php:41-42` leaves both at the driver default:

```php
'busy_timeout' => null,
'journal_mode' => null,
```

SQLite's default busy timeout is **zero** — a writer that finds the database locked fails
immediately rather than waiting. With WAL unset, readers block writers too.

This is `decision/0010`'s pattern precisely: a semantic doccum depends on, inherited from a driver
default instead of stated. The item whose `doneWhen` asked for the cause to be "reproduced or ruled
out ... and stated" closed with it unstated, while the container printed it.

## What the backoff did and did not do

`$backoff = [5, 15]` (#83) is still correct and is probably why this now recovers instead of
failing: a retry 5 seconds later finds the lock gone. But it treats the symptom. The fix is to state
the semantics — a non-zero `busy_timeout` and WAL journalling for the embedded database — so writers
wait rather than fail, and readers stop blocking writers.

## Also worth deciding

Whether queue, cache and session all belong on the same SQLite file as the application data in the
embedded configuration, given four processes contend for it.

Issue #91, opened 2026-09-17
The documented single-container run keeps SQLite inside the image layer, not on /data


abdullah-life:
## Data loss in the documented deployment

CLAUDE.md's Docker section says:

> `/data` is the only persistent volume: SQLite, objects, `runtime.json`, `minio.env`.

**That is false for SQLite**, in the run command CLAUDE.md itself documents:

```
docker run -d -v doccum:/data -p 8080:8080 doccum:local
```

Proof, from the container's own error output on `main`:

```
Connection: sqlite, Database: /var/www/html/database/database.sqlite
```

Inside the image layer. Every user, permission, directory, grant, file row, property and search
index is lost on `docker rm` or any image upgrade — while objects and `minio.env` survive on the
volume, so the instance comes back looking half-alive rather than empty.

## Mechanism

- `config/database.php:38` — `'database' => env('DB_DATABASE', database_path('database.sqlite'))`
- `compose.yaml:13` sets `DB_DATABASE: /data/doccum.sqlite` — **the Dockerfile sets nothing**
- `docker/entrypoint.d/49-doccum-init.sh:15` computes `SQLITE_FILE="${DB_DATABASE:-/data/doccum.sqlite}"`
  and creates it — so `/data/doccum.sqlite` exists, is touched, and is then **never used** by the app
- `RuntimeConfigServiceProvider::applyDatabase()` only acts when `runtime.json` carries a `database`
  block, which the default installer path never writes

So compose users are fine and `docker run` users silently are not.

## Why the smoke does not catch it

The persistence check is:

```
docker restart doccum-smoke
```

`docker restart` keeps the container's writable layer, so a database inside the image layer survives
it. The step passes and proves nothing about the property CLAUDE.md claims.

**The fix to the smoke is one line**: replace `docker restart` with `docker rm -f` plus a fresh
`docker run` against the same named volume. That is what "persists" has to mean for a container
people are told to upgrade by replacing.

## Fix

Set `DB_DATABASE` in the image so the documented run is correct by default, rather than relying on
every operator passing it. Then make the smoke destroy and recreate the container so the claim is
tested rather than asserted.

Found in an adversarial review of the loop's own guarantees; verified against `config/database.php`,
`compose.yaml`, the entrypoint, and the container's own runtime error output.

Issue #57, opened 2026-09-17
ExtractText intermittently fails in the container on a plain .txt upload


abdullah-life:
**Ledger id:** `item/extract-text-flake`
**Size:** S
**Depends on:** nothing

## What happened

The `image` job failed on [#56](https://github.com/turbophp/doccum/pull/56) — a PR whose entire diff is two JSON files under `docs/ledger/` that nothing in the container reads:

```
extraction reached a terminal non-success status (failed) for smoke-35212200191-1.txt
App\Jobs\ExtractText .... FAIL
```

The same job passed on a re-run of the identical commit, and `main` at `3702b8f` — byte-identical application code — passed five minutes earlier. So it is intermittent, not a regression.

## Why this is not "just a flake"

The uploaded file is a plain `.txt`. Extraction needs no `tesseract`, no `pdftotext`, no external binary — it is the simplest path through the ingest pipeline. A terminal `failed` status there is a real defect, and `CLAUDE.md`'s posture is explicit that a failing test is never written off as infrastructure.

The container was healthy when it happened: supervisor showed `minio`, `frankenphp`, `worker` and `worker-ingest` all `RUNNING`, every migration applied, and the installer answered.

Leading hypothesis, **unverified**: a race between the object landing in MinIO and `worker-ingest` picking up the job — the worker reads an object that is not yet readable and the job fails terminally instead of retrying. The container smoke test already had to learn a related lesson once (`ExtractText` sets `file_texts.status = done` and only *then* dispatches `ReindexSearchDocument`, so the smoke test had to wait for the index rather than for extraction — see run/0004).

## What makes this worth fixing before v1

A self-hoster hits this as "I uploaded a document and its text never appeared", with a `failed` row they cannot see and no retry. It is invisible to the test suite because the test queue is synchronous and production is not — which `CLAUDE.md` calls out directly.

## Done when

1. The failure mode is reproduced or the race is ruled out by reading `ExtractText` and the ingest worker — say which.
2. If it is the object-store race: the job retries rather than failing terminally on a transient read, with the retry policy stated and tested through `ProcessRunner::fake()` / a faked storage failure, not by waiting.
3. A terminal extraction failure surfaces somewhere an operator can see it, rather than only in `file_texts.status`.
4. If the cause turns out to be the smoke test racing rather than the app, the smoke test is fixed instead — and the issue says so plainly rather than quietly changing the test.

Found by the container smoke job, which has now caught two defects that six green matrix legs did not.

---
[...]

PR #98, opened 2026-09-17
fix: land the operator in doccum after installing it, not on Laravel's starter page

### LINKED ISSUE #97: Completing the doccum installer lands the operator on Laravel's starter welcome page
## The first thing a new instance shows you is not doccum

`app/Livewire/Setup/FirstRun.php:412`, the last line of the installer:

```php
return redirect('/');
```

`routes/web.php:12`:

```php
Route::view('/', 'welcome')->name('home');
```

`resources/views/welcome.blade.php` is the **stock Laravel starter page** — 202 lines, four links to Laracasts and the Laravel documentation. An operator who has just named their instance, created an administrator and had a home directory provisioned is dropped onto a marketing page for a PHP framework.

`config/fortify.php:91` already says where they should be:

```php
'home' => '/files',
```

So logging in afterwards goes to the right place. Only the one path a brand-new instance is guaranteed to take goes to the wrong one.

## The smoke knows, and steps around it

`.github/scripts/container-smoke.mjs` waits for the welcome page and then navigates away by hand, with a comment explaining why:

```js
page.waitForURL((u) => u.pathname === '/', { timeout: 15000 }),
...
// Checked here rather than on the post-installer redirect: that lands on
// `/`, which is Route::view('/', 'welcome') -- a standalone document that
// does not use the topbar layout, so there would be no topbar to drive.
```

The end-to-end test that exists to drive the product the way an operator meets it documented this in a comment instead of failing on it. That is the same shape as #91 and #92: the check that should have caught it was written to accommodate it.

## Fix

`FirstRun::submit()` should redirect to `route('files.browse')`, matching `config/fortify.php:91`. The smoke should then wait for `/files` and drive the topbar on the page the installer actually delivers, which deletes the workaround and makes the assertion mutation-visible: revert the redirect and the smoke times out.

## Deliberately not in scope

What `/` should render for an authenticated user, and whether `dashboard` (four placeholder boxes today) should live there, belongs to `item/home-dashboard` (order 23). This issue is only about where the installer sends you.

abdullah-life:
Closes #97.

## What an operator saw

`FirstRun::submit()` ended in `return redirect('/')`. `routes/web.php:12` is `Route::view('/', 'welcome')` — the **stock Laravel starter page**, 202 lines with four links to Laracasts and the framework documentation.

So the very first screen after naming your instance, creating an administrator and having a home directory provisioned was a marketing page for a PHP framework.

`config/fortify.php:91` already said where they belong:

```php
'home' => '/files',
```

Every later login went to the right place. Only the one path a brand-new instance is **guaranteed** to take went somewhere else.

## Why the smoke never caught it

Because the smoke was written around it:

```js
page.waitForURL((u) => u.pathname === '/', { timeout: 15000 }),
...
// Checked here rather than on the post-installer redirect: that lands on
// `/`, which is Route::view('/', 'welcome') -- a standalone document that
// does not use the topbar layout, so there would be no topbar to drive.
```

It waited for the wrong page, then navigated to `/files` by hand, and left a comment explaining the defect. The end-to-end test that exists to meet the product the way an operator does **recorded** this in prose instead of failing on it — the same shape as #91 and #92, where the check that should have caught the bug had been shaped to accommodate it.

It now waits for `/files` and drives the topbar on the page the installer actually delivers. There is no `page.goto()` after the redirect any more, so nothing downstream can paper over a landing somewhere else.

## Correction: this assertion *can* be mutation-verified, and will be

An earlier version of this description said mutation-verifying the smoke assertion was impossible here because building a mutated image needs a `composer install` this environment cannot reach. **That was wrong**, and the commit message carries the same error. The convention `decision/0012` established does it in CI: push a branch with the feature deliberately broken and read the `image` job — which is exactly how `claude/mutation-forgot-oracle` was recorded.

The first attempt at that run is void: it was branched from this PR's first head, which was red for an unrelated reason (below), so an `image` failure on it would not have distinguished the mutation from the pre-existing breakage. It will be re-run against a green head and the run recorded here before this merges.

## Two things the first push got wrong

**A fourth assertion site.** `FreshInstallTest` asserted the old redirect and I missed it, because I grepped two named files instead of the tests tree. All six legs caught it. `ProfileUpdateTest`'s `assertRedirect('/')` is deliberately left alone — that is account deletion, and a deleted user belongs at the public root.

**An unreadable diagnostic.** The `image` job failed for an unrelated reason and this is the entire message:

```
timed out waiting for extraction (last status: TINKER_ERROR(Command
```

`last` is taken from `/STATUS:(\S+)/`, which stops at the first space, so `Command failed: docker exec ...` became the single token `TINKER_ERROR(Command` and the reason was discarded. The extraction poller now keeps the thrown message whole and prints it on timeout. That is `decision/0017` one layer down — a diagnostic is worth only what someone can read off it — and it is the second time in two runs that a diagnostic fired and told nobody anything (#57 was the first).

Included here rather than split out because it is the failure this PR hit, in a file this PR already changes, and without it the next `image` failure is equally mute.

## Deliberately not in scope

What `/` should render for an authenticated user, and whether `dashboard` (four placeholder boxes today) belongs there — that is `item/home-dashboard`. `route('home')` remains the brand link on the auth layouts and the post-logout target, both unchanged.


abdullah-life:
## Status: `image` has failed twice here, and I am not calling it a flake

All six test legs are green. The `image` job is the only red check, and it has now failed **twice** on this branch at the same point — the `setup` phase, polling for extraction.

**It is not established as someone else's failure.** `main`'s `image` job is green on its last four commits (`b0f9a62`, `e8686a0`, `8ceaa72`, `12b693d`), so I cannot claim this reproduces on the base. The rules say a re-run only earns "flake" when the failure reproduces off this branch, and it does not.

**But I also cannot yet say what fails**, and that is my own doing twice over.

The first attempt reported:

```
timed out waiting for extraction (last status: TINKER_ERROR(Command
```

truncated at the first space, because `last` comes from `/STATUS:(\S+)/`. I widened it to print the message "in full". The second run then reported, in full:

```
Command failed: docker exec doccum-smoke php artisan tinker --execute
$f = \App\Models\File::where('name', 'smoke-35251931830-1.txt')->first(); ...
```

That is `execFileSync`'s entire `message`: **the command, echoed back.** The cause was never in `message`. It lives on the error object — `status`, `signal`, `killed`, `stderr`. `37c98f9` raises those from `tinker()` itself.

So: twice I made this smoke louder without making it informative. `decision/0017` says a diagnostic is worth only what someone can read off it, and that is not satisfied by printing more characters.

## What is actually known

- `supervisorctl` at failure time: `frankenphp`, `minio`, `scheduler`, `worker`, `worker-ingest` all **RUNNING**, uptime 1m26s. Nothing had crashed.
- The same `tinker()` helper succeeded moments earlier in the same run, for the pragma check added in #94.
- It then failed on **every** poll for the full 60-second window.
- The poller spawns a complete artisan boot every two seconds while extraction is itself running.

Timeout, OOM and a PHP error all fit that shape. One more run distinguishes them, so I am not guessing between them.

## Relation to #92

#94 fixed one cause of the extraction flake — `busy_timeout` at zero — and proved it by mutation. It did not prove that was the *only* cause, and #92's history (a flake that ran roughly 5 in 30) is a reason to expect more than one. If the next run shows a PHP error here, it reopens that question rather than closing it.

The redirect change this PR is actually about is unaffected: the `setup` phase reached the upload, which means it landed on `/files` and drove the topbar there.

---

abdullah-life:
## Narrowed back to the redirect

`f49093c` reverts every smoke change except one assertion. The diff is now 8 lines in `FirstRun.php`, four test assertion sites, and this:

```diff
-      page.waitForURL((u) => u.pathname === '/', { timeout: 15000 }),
+      page.waitForURL((u) => u.pathname === '/files', { timeout: 15000 }),
```

That assertion is undamaged by the revert and is still the whole of this issue's evidence: it runs **before** the `page.goto('/files')` that `main` performs, so putting the redirect back to `/` still times out there. The goto cannot rescue a redirect that went somewhere else.

## Why revert instead of continuing to fix

Five `image` failures, in two distinct groups — and **I caused the second group**:

| runs | failure |
|---|---|
| 1–3 | `STATUS:NO_FILE` — nothing was stored |
| 4–5 | `page.waitForResponse: Timeout 20000ms exceeded` |

Runs 4 and 5 waited on a response to `/livewire/upload-file`. **That endpoint is an assumption I made about how this Livewire version uploads, and never verified.** I cannot verify it here: there is no `vendor/`, and `composer install` is blocked by the environment's network policy, so every smoke change is a speculative push by construction. Instrumentation I could not test became the thing failing — which means runs 4 and 5 say nothing about whether the original defect is still present.

That is the opposite of "one validated push beats three speculative ones", and continuing would have been three more guesses on a PR that is not about the smoke.

## The findings are kept, not dropped

The three defects are real and present on `main`:

1. the upload step asserting page text rather than a stored row,
2. the swallowed `networkidle` race,
3. `exit` making `STATUS:NO_FILE` and `NO_VERSION` unreachable — the status was in `stdout` the whole time, behind a thrown exception.

They are tracked in #106 and move to their own PR, where iteration is the expected cost and where the first question can be the one I skipped: *what does this Livewire version actually do when a file input changes?*

## Still owed here

The smoke mutation run for the `/files` assertion, against a green head, per `decision/0012`. Not claimed until done.

---

abdullah-life:
## Blocked on #107 — and I am stopping pushes here

Sixth `image` failure, back to the unreadable form:

```
extraction for smoke-… did not reach "done" within 60000ms (last status: TINKER_ERROR(Command
```

That is a direct consequence of my own revert: narrowing this PR took the *diagnostics* out along with the speculative fixes, so the failure is illegible again. Making the PR small and making the failure readable were in tension and I traded the wrong one away.

**State of the evidence:**

- 13 of 14 checks green. All six matrix legs pass, so the redirect change and its four assertion sites are proven.
- `image` has failed **6 times** here; `main`'s has passed its last **3** (`e8686a0`, `8ceaa72`, `12b693d`).
- The diff is now one line of smoke plus the redirect, so this is not my instrumentation any more.

**What I am not going to do** is push a seventh guess. Every smoke change I make is unverifiable locally — no `vendor/`, `composer install` blocked by the environment's network policy — so iterating here is guessing with a 3-minute feedback loop, and two of the six failures were caused by exactly that.

**The plan instead:** #107 lands the diagnostics on `main`, where the `image` job is green and they can be judged against a passing baseline. Once they are in, this PR rebases and the failure becomes legible — `STATUS:NO_FILE` with an exit status and stderr, instead of a truncated token.

**What is still not established:** whether landing on `/files` after install genuinely perturbs the upload, or whether this is the long-running extraction fragility (#92 was never proven to have a single cause) finding a branch to be unlucky on. Six-versus-three is suggestive, not conclusive, and I am not going to assert either until the diagnostics say which.

This PR stays open and watched. No further pushes until #107 is in.

---

abdullah-life:
Blocker updated: #107 merged, and this is now blocked on **#110**.

Run 9 reproduced `main`'s exact `/` → `/files` page sequence and **still** stored nothing. That refutes my third explanation for this, after the Livewire-readiness wait (run 8) and the upload-endpoint wait (runs 4–5, which were my own error).

The finding that matters is that it is not this PR's defect at all. The same failure now has a second, independent witness: #108's trash-target upload reports `container says: FILE:no` — nothing stored, on a branch that does **not** touch the installer redirect. Same smoke, same `main`, same container.

What survives every refutation is narrow, and it is all I will claim: the upload stores nothing when little happens between navigating into the directory and `setInputFiles`, and works when three topbar assertions sit in between. That is a correlation. I do not have the mechanism.

#110 stops guessing at it — `uploadAndProveStored()` retries the interaction and logs which attempt succeeded, so the next run distinguishes a transient window from something structural. Once it lands, this PR rebases onto it and the workaround commit here (`83fc1f6`, which reproduces `main`'s page sequence) should come back out, since it was aimed at an explanation that turned out to be wrong.

No further pushes here until #110 is in. The redirect fix itself remains proven by all six matrix legs.

---

Issue #106, opened 2026-09-17
The container smoke's upload step can report success when nothing was stored, and its two most diagnostic statuses were unreachable


abdullah-life:
Three defects in `.github/scripts/container-smoke.mjs`, found by three consecutive `image` failures on #98 that said almost nothing until the diagnostic was fixed twice. All three are present on `main`; they are latent there, not absent.

## 1. The upload step asserts text on a page, not a stored file

```js
await page.getByRole('button', { name: 'Upload', exact: true }).click();
await page.getByText(FILE_NAME, { exact: true }).waitFor({ timeout: 10000 });
console.log('[setup] upload accepted, file listed in the directory');
```

`getByText(FILE_NAME)` matches the **file input's own displayed filename** and Livewire's optimistic preview. Neither requires a `files` row to exist.

It printed *"upload accepted, file listed in the directory"* through an entire run in which `File::where('name', …)->first()` returned **null** for the following sixty seconds. This is `decision/0012` exactly: an assertion that observes a resting state instead of requiring the feature to *do* something.

## 2. A swallowed race is what makes that reachable

```js
await page.locator('input[type="file"]').setInputFiles(tmpFile);
await page.waitForLoadState('networkidle', { timeout: 10000 }).catch(() => {});
await page.getByRole('button', { name: 'Upload', exact: true }).click();
```

Livewire POSTs the file to its temporary-upload endpoint asynchronously. The smoke waits for the *network to go quiet* and then **swallows the timeout** with `.catch(() => {})`. When that wait does not settle, it clicks Upload anyway against a form referencing a temporary file that has not landed. Nothing is stored, and defect 1 ensures nobody notices.

It is timing-sensitive, which is why it correlates with unrelated changes: #98 removes one full page load from the setup path and the failure became reproducible on that branch (3 of 3) while `main` stayed green (4 of 4). The change did not introduce the bug; it changed the timing enough to lose the race consistently.

## 3. `exit` in the tinker snippet made `NO_FILE` and `NO_VERSION` unreachable

```js
"if (!$f) { echo 'STATUS:NO_FILE'; exit; }",
"if (!$v) { echo 'STATUS:NO_VERSION'; exit; }",
```

PsySH leaves the process with **exit status 1** when `--execute` code calls `exit`, so `execFileSync` throws. Both readings were converted into "the tinker call failed" and never reached the poller. The two most diagnostic answers that query can give have been unreachable for as long as they have existed:

```
exit status: 1
signal: (none)
[...]

PR #111, opened 2026-09-17
MUTATION — DO NOT MERGE — empty TrashFile::handle() to prove #108's trash assertion

abdullah-life:
**Deliberately broken. Not for merge. Opened only to obtain an `image` job, and will be closed as soon as it reports.**

`.github/workflows/tests.yml` runs on `pull_request` and on pushes to `main` only, so a bare branch push produces no `image` job. A pull request is therefore the only way to run the container smoke against a mutated image — which is what `decision/0012` requires and what `item/files-actions-ui`'s `doneWhen` asks for.

## The mutation

`app/Actions/Files/TrashFile.php`, body emptied:

```php
public function handle(File $file): void
{
    // nothing
}
```

The Trash button still succeeds and the panel still closes; nothing is soft-deleted.

## What must happen

The `image` job **must fail**, in `checkTrashRemovesFileFromListingAndSearch()`, at:

```js
await page.getByText(TRASH_CHECK_FILE_NAME, { exact: true })
  .waitFor({ state: 'detached', timeout: 10000 });
```

**If it passes, the assertion proves nothing** and #108 does not satisfy its `doneWhen`, however green the PR itself is. That is not hypothetical here: the mutation harness already caught exactly this on #108, where `browser-select-directory-view` kept passing with its guard deleted because `PropertyPanel::mount()` was refusing first. A guard whose test passes without it is worse than none, because it looks protected.

The unit tests are expected to fail too. The smoke assertion is the one under test.

## Where the result goes

Into the docblock of `checkTrashRemovesFileFromListingAndSearch()` on #108, replacing the `<PENDING …>` placeholder the subagent left there and flagged rather than papering over.

---


abdullah-life:
Done its job — closing. **Never for merge.**

The `image` run failed where it had to, and the call log is the evidence rather than the red mark:

```
locator.waitFor: Timeout 10000ms exceeded.
Call log:
  - waiting for getByText('DoccumSmokeTrashTarget.txt', { exact: true }) to be detached
    25 × locator resolved to visible <a …>DoccumSmokeTrashTarget.txt</a>
```

With `TrashFile::handle()` emptied, the run reached the trash step, uploaded and confirmed the target, found it in search, clicked Trash through the panel — and the file **stayed in the listing**. That is #108's assertion catching the break.

Recorded in `checkTrashRemovesFileFromListingAndSearch()`'s docblock on #108, replacing the `<PENDING …>` placeholder: [run 35259216429, job 105330423553](https://github.com/turbophp/doccum/actions/runs/35259216429/job/105330423553).

## Two traps worth recording

**A bare mutation-branch push produces no `image` job at all.** `tests.yml` runs on `pull_request` and pushes to `main` only, so pushing a mutated branch and waiting — which is what `decision/0012` literally describes — yields silence, and silence reads like success. An earlier mutation branch of mine sat that way unnoticed. Hence this draft PR.

**Four notifications told me the job had failed, and none of them was evidence.** The run could equally have died earlier at the upload race (#106), with the trash assertion never executing. "It went red" and "the assertion caught it" are different claims; conflating them is how `browser-select-directory-view` came to carry a mutation entry it hadn't earned.

Incidentally this run also shows the #106 race again — `attempt 1 left no files row` for the first upload, stored on attempt 2 — on a branch that has nothing to do with the installer redirect.

---

Issue #102, opened 2026-09-17
item/files-versions-replace — version history and replace-in-place


abdullah-life:
Slice **20b** of the `item/files-three-pane` split (#100), from #13. Depends on 20a (#101).

`FileDownloadController` serves `currentVersion` only; there is no way to see or fetch an earlier version.

## doneWhen

> The detail panel lists a file's versions newest first with size, uploader and date; Replace uploads through `StoreFileVersion` and the list then shows version 2 under an unchanged name; a per-version download route authorises through `FilePolicy::download` and returns 404 for a version id belonging to a different file, with that cross-file check mutation-checked; the container smoke replaces the uploaded file and asserts a second version row appears, shown failing against an image whose Replace posts to the create path instead.

## Notes

- The cross-file version id is the guard here: a per-version route that trusts the id is an IDOR straight past `directory_access`.
- Replace must add a version, never overwrite one — archived periods are read-only, and `StoreFileVersion` already enforces that.

abdullah-life:
Closed by #117 (`0948c6d`), green on all 14 checks.

## doneWhen, clause by clause

**"The detail panel lists a file's versions newest first with size, uploader and date"**
`Browser::render()` builds `versions` as a query — `orderByDesc('version_number')`, `with('uploader')` — and `browser.blade.php` renders a `data-test="file-version-row"` per version with `Number::fileSize()`, the uploader's name and the date. `FileVersion::uploader()` is new; `uploaded_by` was an FK with no relation, so a template reaching for it rendered the raw integer. Ordering is by `version_number`, the domain key `StoreFileVersion` computes, not `created_at` (second-resolution ties) or `id`. The test uses `assertSeeInOrder`, since a count assertion cannot tell "ordered" from "SQLite happened to return rowid order".

**"Replace uploads through StoreFileVersion and the list then shows version 2 under an unchanged name"**
`Browser::replaceFile()`, with its own `$replacement` property — not `$upload`. Sharing one property would mean choosing a replacement and then clicking the main Upload button creates a new file, which is this item's bug reachable by an ordinary user.

The directory is resolved **from the file**, never from `$this->directory`. That is not defensive: after `moveFile()` the selected file lives elsewhere while `$this->directory` is unchanged and the panel still shows it, so authorising and uploading against the browsed directory would make `StoreFileVersion`'s `(directory_id, name_key)` lookup miss and **create a new file there** — the create path, with no mutation involved.

[...]

PR #118, opened 2026-09-17
MUTATION PROOF — DO NOT MERGE — Replace posts to the create path (#102)

abdullah-life:
**This PR exists only to make CI build a deliberately broken image. It is never merged and will be closed as soon as the `image` job has been read.**

Proves that `checkReplaceAddsASecondVersion()` in `.github/scripts/container-smoke.mjs`, added in #117, is load-bearing. CLAUDE.md: *"A smoke assertion is worth only what a mutation says it is… never cite an unmutated assertion as evidence."*

## The mutation

`Browser::replaceFile()` passes the uploaded file's own client name to `StoreFileVersion` instead of the selected file's name:

```diff
-                $this->selectedFile->name,
+                $this->replacement->getClientOriginalName(),
                 $this->replacement->getMimeType(),
```

That is exactly the failure the item's `doneWhen` names — *"an image whose Replace posts to the create path instead"*. `StoreFileVersion` versions by `(directory_id, name_key)`, so under a different name it takes the **create** path and writes a second `File` row instead of appending a version.

## What must fail

The smoke's checksum triple, in the `image` job:

1. `DoccumSmokeVersionTarget.txt`'s current version should hash to the replacement body — under this mutation it still hashes to the **original**, because nothing ever touched that file's row.
2. version 1's checksum unchanged (this half should still pass — it is the "never overwrites" half).
3. no `File` row named `DoccumSmokeReplacement.txt` — under this mutation one **exists**.

A green `image` job here would mean the assertion proves nothing and #117's smoke clause is unsatisfied.

The unit suite is also expected red — `FileBrowserActionsTest`'s replace cases cover the same line. Only the `image` job's output is being read here.

---


abdullah-life:
Mutation read; closing. The record is now in `checkReplaceAddsASecondVersion()`'s docblock in #117.

**Result — both directions measured against a built image:**

| build | `image` job | outcome |
|---|---|---|
| correct (#117 `a095a34`) | [run 35274858080](https://github.com/turbophp/doccum/actions/runs/35274858080/job/105382869492) | **passes** |
| create-path mutation (this PR, `cb6d8bd`) | [run 35274860882](https://github.com/turbophp/doccum/actions/runs/35274860882/job/105382879408) | **fails** |

The failing run stops at the checksum assertion and names the symptom:

```
[setup] DoccumSmokeVersionTarget.txt stored, confirmed by a files row (attempt 1)
container-smoke (setup) failed: replace did not update the document's own
current version checksum
```

**The first attempt at this proof was worthless, which is why both rows above exist.** That revision failed identically on the correct build and on the mutation — both timing out on `[data-test="file-version-row"].nth(1)` — because it selected the uploaded row without re-navigating, so the main upload form's file input still displayed the filename, `getByText()` did not land on the listing's link, and `selectFile()` never fired. A check that fails both ways proves nothing; `mutation-check.php` asserts both directions for exactly this reason, and this one had to earn it the same way.

The check was also reordered so the evidence claimed to be load-bearing is the evidence that fires: it polls the database and reports the checksum triple before touching the DOM. The rendered-row assertion survives as its own separate claim about the image.

The unit matrix on this branch is red too, as expected — `FileBrowserActionsTest`'s replace cases cover the same line. Only the `image` job was being read here.

This PR is closed and is never merged. The branch `claude/nifty-johnson-3xth3u-versions-mutation` is left in place: branch deletion is blocked from the agent environment this loop runs in (the git proxy drops the connection on a delete refspec), so it needs deleting by hand. Its only commit is the deliberate break described above — do not merge it.

---

Issue #121, opened 2026-09-17
"database is locked" on the jobs table, with busy_timeout already applied


abdullah-life:
Seen in the `image` job on #120 at `db79532` ([run 35279130888](https://github.com/turbophp/doccum/actions/runs/35279130888/job/105396743938)). The same commit's smoke passed on the next run, so this is **intermittent**, not a hard break.

## What happened

The smoke died at extraction of its first upload:

```
container-smoke (setup) failed: extraction status "failed" for smoke-35279130888-1.txt:
SQLSTATE[HY000]: General error: 5 database is locked
(Connection: sqlite, Database: /data/doccum.sqlite,
 SQL: update "jobs" set "reserved_at" = 1789682091, ...)
```

Four lines earlier in the same run:

```
[setup] embedded database: journal_mode=wal, busy_timeout=5000ms
```

So WAL is on and the busy timeout is applied — the configuration `sqlite-busy-timeout-stated` and `sqlite-wal-journal-stated` guarantee is in place — and the lock happened anyway. That mutation entry's own `why` cites this exact error on the jobs table as the thing the stated timeout fixes. It does not fix this case, and the entry is now overclaiming.

## Why busy_timeout does not cover this

`busy_timeout` makes a connection **wait and retry** when it cannot get a lock. It does not apply to the one case SQLite refuses to retry: a **deferred transaction that reads first and writes later**. If another connection has written since the read began, SQLite returns `SQLITE_BUSY` *immediately* and will not invoke the busy handler, because waiting could never succeed — the reading transaction's snapshot is already stale and it must be rolled back.

Laravel's database queue worker has exactly that shape: inside a transaction it `SELECT`s an unreserved job, then `UPDATE`s it to set `reserved_at` — which is the failing statement above. The single container runs **two queue workers plus the scheduler plus FrankenPHP**, all on one file, with queue, cache and session all on the database driver. Two workers reaching for the same job is the textbook trigger.

I'm confident about the SQLite semantics and about the failing statement. The claim that Laravel's SQLite driver opens these as deferred rather than immediate transactions is **inferred from the error, not read from the source** — `vendor/` cannot be installed in the agent environment. Confirming that is step one here.

## Likely fix

If the inference holds, the fix is to begin write transactions with `BEGIN IMMEDIATE` so the write lock is taken up front and `busy_timeout` can do its job. Laravel does not do this for SQLite by default. Options, roughly in order of preference:

[...]

Issue #103, opened 2026-09-17
item/files-list-sort-select — dense list, column sort, multi-select, bulk trash


abdullah-life:
Slice **20c** of the `item/files-three-pane` split (#100), from #13. Depends on 20a (#101).

**This is the slice that carries the real risk of the four.** It is the JavaScript-heavy one (range selection, an id array round-tripping through Livewire), the destructive one (bulk trash), and the one a Blade assertion cannot see at all.

## doneWhen

> The listing is a table with name, owner, modified and size; sort is a whitelisted `orderBy` in the query and an unknown column falls back to name, tested; multi-select supports shift-range and ctrl-toggle; bulk trash re-authorises **every** selected id through `FilePolicy::delete` and trashes **nothing** if any one is refused, tested and mutation-checked; the container smoke uploads three files, selects two, bulk-trashes and asserts the third is still listed — shown failing against an image whose bulk method ignores the selection.

## Why the smoke assertion is phrased that way

The original item 20 said "bulk trash of exactly the selected files". A test written to that sentence passes just as well against an implementation that trashes **everything** — the selected files are, after all, gone. The third file is the only witness that the selection was respected, which is why it is in the `doneWhen` rather than left to whoever writes the test.

Same for the all-or-nothing rule: authorising the first id and then trashing the whole array is the obvious implementation and it is wrong. It needs a test with one refused id in the middle of an otherwise permitted selection.

## Notes

- `composer.json` has `livewire/flux ^2.13`, no flux-pro — `flux:table` and `flux:pagination` do not exist here.
- An unwhitelisted sort column reaching `orderBy` is SQL injection; the fallback is the guard, not a convenience.
- Drag-and-drop upload can fold in here as a JS-only addition to the existing upload, or be deferred by a recorded Decision. It must not be a reason this slice slips.

abdullah-life:
Closed by #120 (`1f96f35`), green on all 14 checks.

## doneWhen, clause by clause

**"The listing is a table with name, owner, modified and size"**
`browser.blade.php` renders a plain `<table data-test="files-table">` — Flux's free tier has no `flux:table`. Owner needed a new `File::creator()` relation (`created_by` was an FK with no relation, so a template would have rendered the raw integer), eager-loaded in `render()`. Size via `Number::fileSize()`.

**"sort is a whitelisted orderBy in the query and an unknown column falls back to name, tested"**
[...]

PR #122, opened 2026-09-17
MUTATION PROOF — DO NOT MERGE — bulk trash ignores the selection (#103)

abdullah-life:
**Exists only to make CI build a deliberately broken image. Never merged; closed as soon as the `image` job has been read.**

Proves `checkBulkTrashLeavesUnselectedFilesAlone()` in #120 is load-bearing. CLAUDE.md: *"A smoke assertion is worth only what a mutation says it is… never cite an unmutated assertion as evidence."*

## The mutation

```diff
 $files = File::query()
     ->where('directory_id', $this->directory->getKey())
-    ->whereIn('id', $this->selectedIds)
     ->get();

-abort_if($files->count() !== count(array_unique($this->selectedIds)), 404);
```

Both lines, deliberately. Dropping only the `whereIn` would make the exact-resolution guard abort before anything was trashed, and the check would fail because *nothing* happened rather than because *everything* did — a different bug wearing the same red mark. Removing both models the doneWhen's named failure exactly: **bulk trash that ignores the selection and takes the whole directory.**

That failure is why the survivor assertion exists. As the item puts it: *"bulk trash of exactly the selected files"* also describes an implementation that trashed all of them, and only the third file tells them apart.

## What must fail, and where

The smoke uploads three files, selects two, and bulk-trashes. Under this mutation the **survivor** is trashed too, so the check must report that by name.

**A red `image` job is not sufficient evidence here.** Per #121, the container intermittently dies earlier with `database is locked` on the jobs table during extraction — that happened on #120's own first run. If this job dies there instead, it proves nothing about the assertion and the run has to be repeated. The other direction is already established: #120 is green on all 14 including `image` at `c1b4bd21`.

The unit matrix is expected red too — `FileBrowserListTest`'s bulk cases cover the same lines. Only `image` is being read.

---


abdullah-life:
Mutation read; closing. The record is now in `checkBulkTrashLeavesUnselectedFilesAlone()`'s docblock in #120.

## Result — both directions measured

| build | `image` | outcome |
|---|---|---|
| correct (#120 `c1b4bd2`) | [run 35279512648](https://github.com/turbophp/doccum/actions/runs/35279512648/job/105397955034) | **passes** |
| bulk trash ignores the selection (this PR, `5632724`) | [run 35279957311](https://github.com/turbophp/doccum/actions/runs/35279957311/job/105399358226) | **fails** |

The failing run reached the assertion and named the symptom:

```
[setup] DoccumSmokeBulkSurvivor.txt was NOT left live by bulkTrash()
        -- raw output: SURVIVOR_LIVE:no V1_TRASHED:yes V2_TRASHED:yes
container-smoke (setup) failed: DoccumSmokeBulkSurvivor.txt is not live
        after a bulk trash that should not have selected it
```

All three files trashed — exactly what "ignores the selection" looks like from outside, and exactly what the survivor assertion exists to catch. Note it is the **survivor** half that fired, not the two "was trashed" halves: those pass identically on both builds, since the mutation trashes the selected files just as thoroughly as a correct build does.

## What only the passing direction proves

Worth stating, because it is easy to assume the failing run carries everything. Under this mutation the selection is **ignored entirely**, so this run says nothing about whether multi-select works. The green run does: the survivor stayed live while exactly the two clicked rows were trashed, which can only happen if the checkbox clicks reached `selectRow()` and left `$selectedIds` holding those two ids.

That is the only evidence anywhere that the Alpine mechanism (`$wire.selectRow()` reading `$event.shiftKey` / `$event.ctrlKey`) functions in the shipped image — the single risk flagged as unverifiable when this item was written. No Blade assertion in the suite can see it.

## Why both rows exist

One alone is not a proof. The replace check one item earlier failed *identically* in both directions on its first attempt and was minutes from being recorded as evidence. And a red `image` job now has to be read for **where** it died, not merely that it did — #121 has the container intermittently dying at extraction with a locked `jobs` table, which happened on #120's own first run.

The unit matrix here is red as expected; `FileBrowserListTest`'s bulk cases cover the mutated lines. Only `image` was being read.

Closed, never merged. The branch `claude/nifty-johnson-3xth3u-list-mutation` needs deleting by hand — branch deletion is blocked from the agent environment this loop runs in (the git proxy drops the connection on a delete refspec). Its only commit is the deliberate break above.

---

PR #123, opened 2026-09-17
ledger: record run/0016 — bulk trash, and a clause that cannot be proved

abdullah-life:
Ledger only. Records the iteration that shipped `item/files-list-sort-select`.

## What it records

- **`item/files-list-sort-select` → Completed**, via #120 (merge `1f96f35`). v1 is **22 of 45**.
- PR nodes for #120 (merged) and #122 (closed — the mutation build).
- 649 tests, 1340 assertions.

## Two decisions

**`decision/0024` — a client-supplied selection must resolve exactly, or nothing happens.** `bulkTrash()` resolved `$selectedIds` unscoped, and `whereIn()` **silently drops** an id matching nothing. A crafted id, or a file another user trashed or moved between the listing rendering and the button being clicked, therefore made the method trash a *subset* of the selection and report success. "Trash exactly what I selected" and "trash whichever of those are still here" are different promises to make to someone deleting documents, and only the first is safe to make silently.

**`decision/0025` — a `doneWhen` clause that cannot be constructed is recorded as unprovable, not worked around.** `FilePolicy::delete()` has no per-file component, so "trashes nothing if any **one** is refused" has no construction. Three ways out; only one is honest. Weakening the resolution to make the case buildable is what was originally done (that's `decision/0024`). Writing a mutation entry claiming the stronger property anyway would be a proof of something that cannot happen — the same fiction caught in `browser-select-directory-view` (run 0014) and `file-policy-replace-refuses-trashed-directory` (run 0015). So the entry claims only the provable all-refused case, and the docblock states the gap.

LOOP.md step 4 has always said a clause that can't be satisfied as written gets said plainly rather than faked. This is the first run that had to use it.

## Not a decision, but the run's most valuable finding

**#121.** A red `image` job on the **correct** build turned out not to be the new check at all — the container died at extraction with `database is locked` on the `jobs` table, four log lines after the same run reported `busy_timeout=5000ms`. The pragma is applied and the lock happens anyway, because `busy_timeout` doesn't cover a deferred read-then-write transaction — which is exactly the shape of Laravel's queue worker reserving a job.

That makes `sqlite-busy-timeout-stated`'s own `why` an overclaim (it cites this precise error as the thing the stated timeout fixes), and it's a product defect: an upload silently loses its extracted text under ordinary two-worker contention, leaving a document search cannot find.

## Verification

`php .github/scripts/validate-ledger.php docs/ledger` → `Ledger is sound: docs/ledger`.

Committed on its own branch, never inside a feature PR, per LOOP.md step 8.

**Merge order:** this will not be merged until `main`'s own push run at `1f96f35` is green. At the time of writing, `ledger` has concluded success there and `tests` is still in progress — so that is a condition on merging, not something already established. (An earlier version of this section stated it as already confirmed; it wasn't, and this corrects it.)


abdullah-life:
`image` failed here, and it is **not this PR's diff** — which is two `.jsonld` files under `docs/ledger/` and cannot affect the container.

It died in the replace smoke check:

```
[setup] DoccumSmokeVersionTarget.txt stored, confirmed by a files row (attempt 1)
[setup] DoccumSmokeVersionTarget.txt's current version checksum is …
container-smoke (setup) failed: replace did not update the document's own current version checksum
```

That check is one I added in #117, and the failure is a flaw in it rather than in the product. It set the replacement file and clicked Replace **once**, then polled for twenty seconds. Replace posts through the same Livewire upload path as any other upload, so it is subject to #106 — an upload issued soon after navigation is silently discarded, no request and no error. `uploadAndProveStored()` retries three times for exactly that reason; this check inherited the bug with nothing to absorb it, and polling cannot rescue a click whose upload was discarded.

**Fix is in #124**, which retries the interaction the way the upload helper does. Not ported into this PR: LOOP.md step 8 keeps the ledger on its own branch, and mixing a smoke fix into it would break that.

This PR is unchanged and stays ledger-only. It will be rebased and merged once #124 lands. `main` itself is green at `1f96f35` — this is a flaky check on `main`, not a broken `main`.

---

PR #129, opened 2026-09-18
Fix issue #106: the Upload click was racing the file input's own upload

### LINKED ISSUE #106: The container smoke's upload step can report success when nothing was stored, and its two most diagnostic statuses were unreachable

abdullah-life:
Opened as instrumentation only. The instrument answered the question, so this is now the fix as well.

**Two corrections are posted as comments rather than edited in here:** #106 was already closed by #107, so `Closes #106` links rather than closes; and **#106 §2 had already hypothesised this race** eight runs ago, filing it as the smoke's race with a test-side fix. Read that comment alongside this body.

## What the probe showed

Round one filtered requests to a guessed `/livewire/` prefix and printed **nothing** across a run in which the bug struck three times — the exact mistake `run/0014` already records. Round two filtered nothing but static assets, and the timestamps named the cause in a single run ([job 105432022684](https://github.com/turbophp/doccum/actions/runs/35290482556)):

```
discarded   POST .../upload-file        request at t+0
            POST .../update  (store)    request at t+14ms     <-- races
            no files row                at t+565ms
            200  .../upload-file        at t+570ms

stored      POST .../upload-file        request at t+0
            200  .../upload-file        at t+18ms
            POST .../update  (store)    request at t+24ms     <-- after
```

All three discards had the upload endpoint answer in ~550ms. Every success had it answer in ~18ms. Nothing else separates them.

## Against the decision table this PR opened with

| observation | diagnosis | actual |
|---|---|---|
| no request appears at all | never sent | no — it was sent |
| a request appears and fails, or answers 4xx/5xx | sent and failed | no — it answered 200 |
| a request appears, answers 2xx, and no row exists | state lost | **yes**, and the mechanism is a race |

A file input posts its bytes the moment it changes, and the component's property is not populated until that POST answers. Clicking Upload inside that window dispatches `store()` against an empty property: it fails `required|file`, and the `_finishUpload` commit landing a moment later re-renders over the validation error. Nothing stored, nothing said — and a retry "works" only because by then the upload has landed, which is precisely why three attempts hid it for three runs.

`replaceFile()` posts through the same path and had the same race. Those two forms are the only file inputs in the application (`grep 'type="file"' resources/views`), so there is no third surface carrying this.

## The fix

Both submit buttons get `wire:loading.attr="disabled"` with `wire:target` naming the **upload property**, not a method, so Livewire holds them disabled for the whole upload and the racing click cannot be made.

## Evidence

`checkUploadButtonIsDisabledWhileTheFileIsStillUploading()` stalls the upload endpoint with a Playwright route (matched by suffix, never by a guessed prefix) and inspects the button while it hangs — the real window is about eighteen milliseconds wide, far too narrow to observe by timing.

The **disabled** assertion is the load-bearing one. The enabled-before and re-enabled-after assertions are labelled in the source as controls, not evidence: a resting state proves nothing, and a guard that wedged the button shut forever would satisfy the disabled half on its own.

**Mutation record is posted as a comment on this PR** and recorded in the check's own docblock, per this file's convention. Both directions measured on built images: correct build green, one-attribute-deleted build red at that assertion and nowhere else. `starts enabled` printed on the mutated image too, so only the disabled half carries weight.

That mutation also settles the Flux forwarding question that made `data-test` unusable on `flux:input`: deleting these attributes changed the shipped image's behaviour, so Flux does forward them to the real `<button>` — measured, not assumed.

## Both retries are removed

`uploadAndProveStored()` and `checkReplaceAddsASecondVersion()` each retried the whole interaction three times. That scaffolding is what hid the race, and keeping it would mean the smoke can never report the fix regressing. Both are single-attempt now, and the green build stored all seven uploads on the first try with no `attempt N` anywhere in the log. The replace check's `max(version_number)` assertion tightens from `>= 2` to exactly `2` for the same reason — the loose bound existed only to tolerate a retry landing a third version.

If the guard is wrong or incomplete, this PR's `image` job goes red rather than quietly passing. That is the intended coupling.

## The probe stays

A few hundred log lines that only matter when something is wrong, which is exactly when they are wanted. Its docblock now records the answer rather than the three candidate diagnoses.

Ledger: `item/upload-silent-discard` (order 25), `decision/0030`.


abdullah-life:
## Mutation record

Both directions measured against built images, not reasoned about.

**Correct build** — `image` **success**, [job 105433709149](https://github.com/turbophp/doccum/actions/runs/35291038177/job/105433709149):

```
[setup] stalling the upload endpoint to check the Upload button is disabled in flight (issue #106)
[setup] the Upload button starts enabled (control, not evidence)
[setup] the Upload button is disabled while the upload is in flight -- issue #106's race cannot be clicked
[setup] the Upload button re-enables once the upload lands (control, not evidence)
```

**Mutated build** — one attribute deleted (`wire:loading.attr="disabled" wire:target="upload"` off the Upload button, Replace's guard left intact), #130, `image` **failure**, [job 105433912606](https://github.com/turbophp/doccum/actions/runs/35291156937/job/105433912606):

```
[setup] the Upload button starts enabled (control, not evidence)
##[error]container-smoke (setup) failed: the Upload button stayed ENABLED while the upload
endpoint was stalled -- issue #106 is reachable again: either wire:loading.attr="disabled"
wire:target="upload" is gone from browser.blade.php, or Flux is no longer forwarding those
attributes to the real <button>
```

It fails at that assertion and nowhere else, so the attribute is what the check is measuring. This is also the answer to the forwarding question that made `data-test` unusable on `flux:input`: deleting these attributes changed the shipped image's behaviour, so **Flux does forward them to the real `<button>`** — measured, not assumed.

Note which half carried the weight. `starts enabled` printed on the mutated image too — it is a control and is labelled as one in the source. Only the disabled assertion moved.

## The retries are gone and the uploads landed anyway

The stronger evidence is what the correct build's log no longer contains. Six uploads, every one on its first attempt, no `attempt N` suffix anywhere and no `attempt 1 left no files row`:

```
[setup] smoke-...-1.txt stored, confirmed by a files row
[setup] DoccumSmokeTrashTarget.txt stored, confirmed by a files row
[setup] DoccumSmokeVersionTarget.txt stored, confirmed by a files row
[setup] DoccumSmokeDownloadTarget.txt stored, confirmed by a files row
[setup] DoccumSmokeBulkSurvivor.txt stored, confirmed by a files row
[setup] DoccumSmokeBulkVictimOne.txt stored, confirmed by a files row
[setup] DoccumSmokeBulkVictimTwo.txt stored, confirmed by a files row
```

The run that diagnosed this had three of those same uploads fail on attempt 1. The retry scaffolding that hid it for three runs is deleted, and the suite passed without it.

`max(version_number) = 2` exactly, as the tightened assertion now requires.

#130 is closed unmerged.

---

abdullah-life:
## Correction to this PR's framing

Two things I had wrong when I opened this, both worth stating rather than quietly editing.

**#106 is already closed**, by #107, and its scope was three defects in `container-smoke.mjs`. The `Closes #106` on this branch links rather than closes. The product defect fixed here has no issue of its own; it is `item/upload-silent-discard`, order 25 in the ledger, and `decision/0030` is where it is written down.

**#106 §2 already hypothesised this race**, in these words:

> Livewire POSTs the file to its temporary-upload endpoint asynchronously. […] When that wait does not settle, it clicks Upload anyway against a form referencing a temporary file that has not landed. Nothing is stored […]

That is the mechanism this PR measured, written down eight runs ago. So the probe did not discover the race. What it established is the **address**: #106 filed it as the smoke's race and prescribed a test-side fix — wait for the upload response instead of `networkidle` — so the fix landed in the harness and the product kept the defect. Then the loop spent three runs describing it as "no request, no error" and walking away from an answer it had already written down.

A person choosing a file and clicking Upload inside half a second hits exactly this, and on a slow connection the window is seconds wide rather than the ~18ms it is in CI. Nobody asked whether a user could reach it, because the race had been observed through a test and was therefore attributed to the test.

The measurement is still what settled it, and the mutation record above still stands. But "we found a race" would be a better story than the true one, and the true one is more useful: a race observed through a test gets blamed on the test.

---

PR #131, opened 2026-09-18
Prove object storage persisted, instead of saying so from a SQLite row

abdullah-life:
The verify phase exists to make one claim: `/data` is the only persistent volume, and everything on it survives the container being replaced. It was asserting that claim from evidence that cannot see it.

```js
await page.getByText(FILE_NAME, { exact: true }).waitFor({ timeout: 10000 });
console.log('[verify] the uploaded file is still listed -- object storage persisted');
```

That row comes from `/data/doccum.sqlite`. So does the search hit on the next line. `checkDownloadReturnsTheUploadedBytes()` is the only thing in the file that fetches a byte, and [it runs in setup only](https://github.com/turbophp/doccum/blob/main/.github/scripts/container-smoke.mjs). **Nothing in the verify phase has ever touched `/data/objects`.**

So a replacement container that came back with `/data/objects` emptied, or `minio.env` lost, or the bucket gone (`DocumentStorage::ensureBucket()` is lazy) passes this phase green — while printing the opposite of what it measured.

`Dockerfile:85` records the **mirror** of that failure: database in the image layer, objects on the volume, "half-alive". Verify caught that one, precisely because it reads the database. The other direction has been open the whole time.

This is CLAUDE.md's topbar lesson again. The log line and the assertion had drifted far enough apart that the line named a subsystem the assertion never reached, and nobody noticed because it was green.

## The change

`runVerify()` follows the Download link for `FILE_NAME` and compares a sha256. The body is deterministic from the environment — `runSetup()` uploads exactly that string — so the checksum is computable in verify without carrying state between phases, which is the constraint this file is built around.

The fetch-and-compare half of the setup check is extracted as `downloadAndCompareBytes()` and shared. Neither phase changes behaviour otherwise. The old line stays, corrected to `the DATABASE persisted` — which is what it always measured.

## Mutation

Not an application mutation; the assertion is about the volume, so the mutation is too. Between the two `docker run`s:

```
docker run --rm -v doccum-ci:/data alpine rm -rf /data/objects
```

Expected: the listing and the search assertions stay **green** (they read SQLite, which is untouched) and the new download assertion goes **red**. That split is the whole point — it demonstrates both that the new check sees object storage and that the old ones never did.

Record to follow on this PR before merge.

Found by the design consult on run/0017's backlog review.

---


abdullah-life:
## Mutation record

#132 empties `/data/objects` between the two containers, leaving SQLite intact. [Job 105438040738](https://github.com/turbophp/doccum/actions/runs/35292462372/job/105438040738):

```
[verify] logged in -- the admin account persisted
[verify] the uploaded file is still listed -- the DATABASE persisted
[verify] search still finds it -- the search index persisted -- OK
##[error][verify] Download for smoke-...-1.txt answered HTTP 500 -- the file row exists
         but its BYTES did not come back
```

**All three pre-existing assertions passed on a container with no object storage whatsoever.** That is the finding, not a side effect of it: login, the file listing and the search hit are every persistence claim this phase made before today, and none of them can see `/data/objects`. Only the new assertion reddens.

Correct build: `image` green, with `[verify] the bytes came back from the replacement container -- OBJECT STORAGE persisted`.

## Two rounds of this were my own cost, and worth naming

The **first** mutation round already gave the right answer — the listing assertion passed, the new one failed. But I had placed the bytes check between the listing and the search, so the search assertion never executed and only one of the two old assertions was demonstrated blind. Reordering it last is what produced the table above.

That reorder then broke the fix build, because `downloadAndCompareBytes()` reads the row's own Download href rather than building a URL from an id, and the search assertion leaves the browser on `/search`, which has no file rows:

```
locator.waitFor: Timeout 10000ms exceeded.
  waiting for locator('tr[data-test="file-row"]')
    .filter({ hasText: 'smoke-...-1.txt' }) to be visible
```

Diagnosed from that log rather than guessed at, and the navigation is restored with a comment saying why it is load-bearing. Two CI rounds spent trading a correct-but-partial result for a complete one. I think the trade was right — the claim is that *neither* old assertion can see object storage, and half a demonstration would have been cited as a whole one — but the cost was self-inflicted and belongs in the record.

## One product finding, filed not fixed

The 500 in that log is real behaviour in a state operators can reach (a restore that brought back the database but not the objects). #134 filed. This PR asserts the bytes come back and deliberately says nothing about what should happen when they cannot.

---

PR #132, opened 2026-09-18
MUTATION — DO NOT MERGE: empty /data/objects between the two containers

abdullah-life:
Not for merge. #131's mutation.

Not an application mutation — the assertion under test is about the volume, so the mutation is too. One step added to `tests.yml`, between the two `docker run`s:

```
docker run --rm -v doccum-ci:/data alpine rm -rf /data/objects
```

The verify phase then meets a container whose file rows all survive and whose bytes are all gone.

**Expected**, and the split is the whole point:

| assertion | reads | expected |
|---|---|---|
| `the uploaded file is still listed` | SQLite | **green** |
| `search still finds it` | SQLite | **green** |
| `the bytes came back … OBJECT STORAGE persisted` | `/data/objects` | **RED** |

If all three go red, the mutation was too blunt and proves less than it looks like. If all three stay green, #131's new assertion is worthless and this is how we find out.

Closed and left unmerged once the result is recorded on #131.

---


abdullah-life:
Result recorded on #131. Round three gave the full split: with `/data/objects` destroyed, login, the file listing and the search hit all stayed green and only the new bytes assertion reddened — so none of the three pre-existing persistence assertions could see object storage. Closed unmerged, as intended.

Branch deletion is not available from this environment, so `claude/nifty-johnson-3xth3u-verify-objects-mutation` still exists on the remote and needs deleting by hand.

---

PR #135, opened 2026-09-18
item/file-policy-trashed-file-guards: state every ability's posture on a file trashed on its own (#116)

### LINKED ISSUE #116: FilePolicy::update() may allow editing an independently trashed file
Raised by the design consult on #102, where the same hole is being closed for the new `replace` ability. `update()` is out of that item's scope.

## The gap

`FilePolicy::update()` resolves the file's **directory** through `liveDirectory()` and refuses when the directory is part of a cascade-trashed subtree (that is #62's work, and it is mutation-checked as `file-policy-update-refuses-trashed-directory`).

It does not check whether the **file itself** is trashed. A file soft-deleted on its own, sitting in a perfectly live directory, resolves a live directory, passes `DirectoryAccess::can(..., Edit)`, and `update()` returns `true`.

## Why that is reachable

Livewire re-hydrates a public model property on every request. If its restoration query does not apply the `SoftDeletingScope`, a `$selectedFile` trashed by another user between two requests comes back as a trashed model, and `Browser::renameFile()` would then rename a file in the trash.

**This has not been confirmed against Livewire's actual restoration query** — `vendor/` is not installable in the agent environment, so it was not read. Confirming it is the first step here, not an assumption to build on. The test is cheap: select a file, trash it out of band, call `renameFile`, and see what the component does.

## If confirmed

`if ($file->trashed()) { return false; }` at the top of `update()`, a `FilePolicyTest` case, and a `.github/mutations.json` entry. Consider whether `move()` and `legalHold()` want the same thing — `delete()`, `restore()` and `purge()` deliberately do not, since those are how a trashed file gets out of the trash.

## If not confirmed

Close it, and say in the close comment what the restoration query actually does — that is the fact worth keeping, and it is the same fact #102's `replace` guard rests on.

---

## Resolved by #135 (`item/file-policy-trashed-file-guards`, order 26)

**CONFIRMED — and the answer is the "if confirmed" branch.** The reachability test asserted a *specific* status rather than "some error", because 403 and 404 mean opposite things here, and CI answered green on the first run across five test jobs (PHP 8.4/8.5 × sqlite/pgsql/mysql).

**Livewire's restoration does NOT apply the `SoftDeletingScope`.** It hands `$selectedFile` back as the trashed model, `renameFile()` reaches `authorize('update', ...)`, and the policy is what refuses. So this was reachable, not theoretical — and `replace()`'s guard rested correctly on that same unread fact. Both the implementing worker and the design consult had reasoned toward 404, from the implicit-route-model-binding analogy and `newQueryForRestoration()`; the measurement overturned both.

That fact is recorded in the test's own comment in `tests/Feature/FileBrowserActionsTest.php`, where the next reader meets it, including what a future red would mean (a Livewire upgrade changing the answer, not a broken test).

### Each `doneWhen` clause

- **Every ability states its posture** — `FilePolicy`'s class docblock now documents all three groups: `replace()`/`update()`/`move()`/`legalHold()` refuse; `delete()`/`restore()`/`purge()` deliberately never check `trashed()`, because they are how a file gets into the trash, out of it, or destroyed.
- **`update()`, `move()`, `legalHold()` refuse a trashed file** — done, each placed as `replace()` places its own.
- **A `FilePolicyTest` case covers each** — three cases, each trashing the file *alone* with its directory left live.
- **Each carries a `mutations.json` entry** — `file-policy-{update,move,legal-hold}-refuses-trashed-file`, all three confirmed load-bearing by the `guards` job.
- **The reachability question settled, not assumed** — above.

### One thing the item did not anticipate

The `guards` job failed the first round, on three **pre-existing** mutations rather than the new ones. `TrashDirectory` cascades onto files, so the new trashed-*file* guard runs before `liveDirectory()` and refuses on its own — and `file-policy-{update,move,legal-hold}-refuses-trashed-directory` stopped being load-bearing, with no change to their own code and the suite fully green.

Adding a correct guard can silently de-fang an older one's proof. `replace()` hit the identical trap when its guard was added; those three tests now carry the same isolation — restore the file, leave the directory trashed, and assert `$live->trashed()` is false first so the premise cannot rot silently.

abdullah-life:
Item 26. `FilePolicy` had three postures on a file soft-deleted **on its own** while its directory stays live, and only one of them was deliberate:

| ability | before | after |
|---|---|---|
| `replace()` | refuses | unchanged |
| `update()`, `move()`, `legalHold()` | **pass** — only ever asked `liveDirectory()`'s question | refuse |
| `delete()`, `restore()`, `purge()` | pass, deliberately | unchanged, now documented |

The third group must never check `trashed()` — those are how a file gets into the trash, out of it, or destroyed. That was already true and nowhere written down, which is half of what made the second group's omission invisible.

## The reachability question, and how this settles it rather than assuming it

Whether this is reachable through the browser depends on a fact about Livewire that **cannot be read in the agent environment** — there is no `vendor/`. The item exists partly because that fact was assumed once already, by `replace()`'s guard.

So the Livewire test is split in two:

- **The guarantee** — a file trashed on its own is not renameable through the browser. True under *both* mechanisms, asserted without reference to which guard produced it, green either way. This is the security property the item owes.
- **The discriminant** — asserts the one thing that differs:
  - **403** → Livewire hands the trashed model back, execution reaches `authorize('update', ...)`, and the new guard refuses.
  - **404** → restoration re-queries through the default scope, `$selectedFile` is null, and `renameFile()`'s own `abort_if` answers first — so the new guard is unreachable *through this surface*, though it still answers every direct `can('update', ...)` call.

**The discriminant's assertion is a bet, and its comment says so.** A red reading `expected 403, got 404` is the test doing its job and *is* the answer — flip it, record it, don't reason around it. Neither outcome changes the guarantee or whether the policy guards are correct. `assertForbidden()`-or-404 would have passed in both worlds and settled nothing, which is the over-determination failure behind all three of this loop's earlier false mutation proofs.

## Mutations

Three new entries, each `expectFailing` pointing at a **direct** `FilePolicyTest` case rather than the Livewire one — so the mutation verdicts don't depend on the unresolved question above.

One real catch from the worker worth recording: a bare `if ($file->trashed()) { return false; }` block copied into all three methods would have made that literal string appear **four** times in `FilePolicy.php`, silently breaking the *existing* `file-policy-replace-refuses-trashed-file` entry's uniqueness. The per-method comment wording is what keeps every entry unique. Verified here against `mutation-check.php`'s own `substr_count()` logic: **all 29 entries match exactly once**, the seven pre-existing FilePolicy ones included.

The policy tests trash the file alone and leave the directory live — the cascade trap that let an earlier `replace()` test pass with its guard deleted.

## Verified / not verified

Verified locally: `php -l` on every changed file; the 29-entry uniqueness check; each `expectFailing` matching one `it()` description.

**Not verifiable here:** the suite, pint and phpstan cannot run without `vendor/`, and Livewire's hydration source cannot be read. CI is the verification of record.

Closes #116

---


abdullah-life:
## The discriminant answered: 403, and #116 was reachable

CI settled it green on the first run, across five test jobs (PHP 8.4/8.5 × sqlite/pgsql/mysql). **Livewire's restoration does not apply the `SoftDeletingScope`** — it hands `$selectedFile` back as the trashed model, `renameFile()` reaches `authorize('update', ...)`, and the new `update()` guard is what refuses.

Two things follow:

1. **#116 was reachable, not theoretical.** A file trashed by another user between two of this user's requests really would have been renameable.
2. **`replace()`'s own guard was correctly placed** on this same unread fact.

Both the implementing worker and the design consult argued for 404 — from the implicit-route-model-binding analogy and from `newQueryForRestoration()`. The measurement says otherwise. That's the third time in this loop a plausible chain of reasoning about framework internals has been overturned by one CI run, which is why the item was written to measure rather than argue.

The test comment said *"THIS IS A BET, NOT A FACT"*. It now records the answer instead — leaving the old wording would be exactly the stale claim-in-a-comment that cost three runs on #106.

## The `guards` job caught something the suite never would have

First round, `guards` failed — not on the new guards, which all passed their own mutations, but on three **pre-existing** ones:

```
file-policy-update-refuses-trashed-directory      STILL PASSES with the guard deleted
file-policy-move-refuses-trashed-directory        STILL PASSES with the guard deleted
file-policy-legal-hold-refuses-trashed-directory  STILL PASSES with the guard deleted
```

`TrashDirectory` cascades onto files, so a cascade leaves the *file* trashed too. The new guard runs first and refuses on its own, so deleting the directory guard stopped changing the outcome.

Worth naming the shape: **adding a correct guard silently de-fanged three tests that were genuinely load-bearing the day before**, with no change to their own code. The suite stayed green throughout. Only the mutation harness could see it.

`replace()` hit the identical trap when its guard was added and already carried the fix, so the same isolation is applied: restore the file, leave the directory trashed, and assert `$live->trashed()` is false first so the premise cannot rot silently. It is a reachable state, not a contrivance — `restore()`/`delete()`/`purge()` resolve the directory `withTrashed()` precisely so a cascade-trashed file can be restored one at a time.

All 29 mutation entries re-verified as matching exactly once.

---

PR #139, opened 2026-09-18
item/files-three-pane: a tree rooted at the viewer's reach, and a real breadcrumb (#104, #99)

### LINKED ISSUE #99: A grant on a nested directory is unreachable from /files: the browser lists filesystem roots, not the viewer's reach
## Resolved by #139 (`item/files-three-pane`, order 27)

`Browser::render()` filtered the centre pane on `where('parent_id', $this->directory?->getKey())`, which with nothing selected resolves to `parent_id IS NULL`. A directory granted directly to a viewer, with no grant on any ancestor, has a **non-null** `parent_id` — so it never appeared at `/files` and was reachable only by typing its URL.

The filtering was never a leak; the defect was reachability.

Both the sidebar and the landing pane now list the viewer's **reach roots** — viewable directories whose parent is not viewable, including those whose parent is null. Resolution lives in `DirectoryAccess::reachRootIds()` / `reachTree()`, built from the existing `viewableDirectoryIds()` seam with one extra query for the candidate set's `parent_id`, checked in memory — no per-directory `can()` loop, and nothing in Blade decides visibility.

**The failing-on-main test** is `'reaches a directory granted directly on a nested node from the landing pane, not only by its own URL'`: `UngrantedParent` → `NestedReachRoot` with a View grant on the child only, then `assertSee('NestedReachRoot')` on the landing page. On main the `where()` clause excludes the row before `whereIn($viewable)` is consulted, and with no directory selected the breadcrumb is empty too, so the name appears nowhere in the response.

## One thing worth recording

Closing this required widening `selectDirectory()`'s lookup as well, and the first attempt got it wrong in an instructive way: scoping the root branch to `reachRootIds()` *alone* meant an id the viewer cannot view failed to **resolve**, so `findOrFail()` answered 404 before `authorize()` was reached — silently replacing a policy decision with an existence check. All six test jobs caught it via `'refuses to select a directory the viewer cannot view'`.

The lookup is now a **union** of true filesystem roots and reach roots, so it only ever widens what can be found and authorisation stays the gate. Which way that *should* answer — 403 or 404 — is #109's question for the whole surface, and deliberately not settled inside a UI item.

### LINKED ISSUE #104: item/files-tree-sidebar — a tree rooted at the viewer's reach, and a real breadcrumb
### Each `doneWhen` clause

- **Sidebar renders Home first, then reach roots and their viewable descendants, asserted against the query** — `DirectoryAccess::reachTree()` returns the roots with viewable descendants nested via `children`, alphabetical at each level, built from `viewableDirectoryIds()` plus one `parent_id` query. `Browser::render()` calls it once and feeds the same collection to the sidebar and the landing pane. `DirectoryAccessTest` covers nested-grant-is-a-root, root-level grants, an admin's true-roots-only case, empty grants, and nesting.
- **A grant on a nested directory alone makes it reachable from `/files`, as a test failing on main** — `'reaches a directory granted directly on a nested node from the landing pane, not only by its own URL'`. See #99 for exactly why it fails on main.
- **Breadcrumb lists every viewable ancestor and no others** — `breadcrumbTrail()` walks `ancestorIds()` narrowed by `$viewable`, ordered by depth.
- **Collapse state per-browser** — Alpine + `localStorage`, read and write in try/catch, tree renders correctly when it comes back empty.
- **Container smoke navigates two levels and asserts three breadcrumb items** — `checkBreadcrumbNavigatesTwoLevels()`, mutation-proven both directions (correct green, ancestor-walk collapsed → `got 1`).
- **`FileBrowserTest` keeps its unreachable-directory guarantees** — untouched; `'Theirs'`/`'Hidden.pdf'` are fully ungranted and unaffected.

### Two things this item changed that were not in its own scope

**A move-destination test had to change what it asks.** `'lists only edit-reachable directories as move destinations'` asserted `assertDontSee('ViewOnlyDestination')` over the whole page, and the sidebar now renders every viewable directory on every `/files` page — which is this item. No fixture can dodge it and page-wide `assertDontSee` cannot be scoped to the dropdown, so it now asks whether a View-level destination can actually be **moved into**, which is the property the narrowing exists to protect. Stated plainly: the UX-level claim that the dropdown omits the option is no longer covered; the security property is the refusal, already covered server-side and now on the tighter View/Edit boundary case too.

**Thirteen existing Playwright locators needed rescoping.** The sidebar renders the same directory name as the centre pane, so every bare `getByRole('link', { name: ADMIN_USERNAME })` became ambiguous under strict mode. Mechanical, verified: of 14 removed lines, 13 are those locators and one is a stale comment.

### Known, filed separately

Two links now read "Home" with different destinations (#140) — a real accessibility wrinkle, and a wording decision against spec §10 rather than a one-line fix.

abdullah-life:
Item 27. The Files surface gains a sidebar tree rooted at the viewer's **reach roots** — viewable directories whose parent is *not* viewable — and the landing pane at `/files` now lists the same set instead of true filesystem roots.

## The bug this closes (#99)

`Browser::render()` filtered the centre pane on `where('parent_id', $this->directory?->getKey())`, which at the root is `parent_id IS NULL`. A directory granted directly to a viewer, with no grant on any ancestor, has a non-null `parent_id` — so it never appeared at `/files` and was reachable only by typing its URL.

The failing-on-main test is `'reaches a directory granted directly on a nested node from the landing pane, not only by its own URL'`: `UngrantedParent` → `NestedReachRoot` (View grant on the child only), then `assertSee('NestedReachRoot')` on the landing page. On main the `where()` clause excludes the row before `whereIn($viewable)` is consulted, and with no directory selected the breadcrumb is empty too, so the name appears nowhere.

## The seam

Reach-root resolution lives entirely in `DirectoryAccess::reachRootIds()` / `reachTree()`, built from the existing `viewableDirectoryIds()` plus one query for the candidate set's `parent_id` column, checked in memory — no per-directory `can()` loop. `Browser::render()` calls `reachTree()` once and feeds the same collection to both the sidebar and the landing pane. Nothing in Blade decides visibility.

`selectDirectory()` now resolves against `reachRootIds()` when browsing the root, because a nested-only reach root's `parent_id` isn't null and the old lookup would 404 on exactly the row the landing pane now shows.

## A pre-existing test this item legitimately invalidates

The worker flagged this and did **not** edit it, which is what I asked for. `'lists only edit-reachable directories as move destinations'` asserted `assertDontSee('ViewOnlyDestination')` over the whole page — and the sidebar now renders every viewable directory on every `/files` page. That *is* the item.

No fixture can dodge it: any viewable directory appears in the tree, as a reach root or as a viewable descendant. And page-wide `assertDontSee` cannot be scoped to the dropdown.

So it now asks the stronger question — **can a View-level destination actually be moved into?** That is what the narrowing exists to prevent, no page text can fake it, and it sits exactly on the View/Edit boundary, where `'refuses to move a file without edit access on both ends'` uses a destination the viewer cannot see at all.

**Stated plainly rather than buried:** what is no longer covered is the UX-level claim that the dropdown omits the option. The security property is the refusal; it was already covered server-side and is now covered on the tighter case too.

## Smoke

`checkBreadcrumbNavigatesTwoLevels()` creates a subdirectory, navigates in, creates a second inside it, navigates in, requires exactly 3 `[data-test="breadcrumb-item"]`, then returns home via the sidebar's own tree link (proving that link is real, not merely rendered).

All 13 pre-existing `getByRole('link', { name: ADMIN_USERNAME })` calls are rescoped to `[data-test="directories-list"]` — **necessary, not cosmetic**: the sidebar renders that same directory name on every page, so every one of those locators would have tripped Playwright's strict-mode uniqueness check. I verified the change is mechanical: of 14 removed lines, 13 are those locators and the fourteenth is one stale comment the worker updated.

Mutation record to follow — the predicted load-bearing assertion is the `count !== 3` breadcrumb check, and it is a prediction, not a claim.

## Known, out of scope

Two links now read "Home" on one page — the topbar's and the sidebar's — with different destinations. A `data-test="sidebar-home"` hook exists so locators can disambiguate, but it is a real accessibility wrinkle for screen-reader users. Filing separately rather than widening this PR.

Closes #104
Closes #99

---


abdullah-life:
## Mutation record

Both directions on built images. #141 collapses `breadcrumbTrail()`'s ancestor walk to the current directory alone — one line, nothing else touched.

**Correct build** — `image` **success**, [job 105451018651](https://github.com/turbophp/doccum/actions/runs/35296836432/job/105451018651):

```
[setup] checking the sidebar/breadcrumb tree navigation two levels deep (issue #104/#99)
[setup] breadcrumb shows all 3 ancestors two levels into the tree
```

**Mutated build** — `image` **failure**, [job 105451024852](https://github.com/turbophp/doccum/actions/runs/35296836432/job/105451024852):

```
expected 3 breadcrumb items two levels into the tree
(root, DoccumSmokeTreeLevel1-…, DoccumSmokeTreeLevel2-…), got 1
```

The `count !== 3` assertion is load-bearing, **measured rather than predicted** — and it took four rounds to earn that, because the first three proved nothing and I want the reason on the record.

## The check was broken three separate ways, and each failure was informative

**Round 1–2 — it failed identically on the correct and mutated builds.** Deterministically, three runs, same locator. A mutation that reddens both directions distinguishes nothing, so no evidence was recorded. Two of my own diagnoses were wrong before measuring: a stale `DirectoryAccess` cache (disproved — `Directory::saved($flushAccess)` is registered in the provider) and a flake (disproved — three identical deterministic failures).

**The real cause:** after clicking into a subdirectory the check waited for the `New folder` field, which exists on *every* directory page. The wait was satisfied by the page being **left**, before `wire:navigate` swapped the DOM, so `fill()` wrote into a doomed field and the subdirectory was never created. CLAUDE.md's topbar lesson in different clothes — a signal already true before the action proves nothing about the action. The comment I replaced even claimed it waited on "the NEXT page's own content".

**Round 3 — the failure moved, and lied about someone else's code.** With navigation fixed, the check passed and then the *upload-guard* check failed with `issue #106 is reachable again`. That guard is present and correct at `browser.blade.php:223-224`. The breadcrumb check's **final** wait was the same non-discriminating kind, so it returned mid-navigation; the upload check began on the outgoing page, its upload fired, and `wire:navigate` swapped in a fresh button with no upload in flight, which reads enabled.

A non-discriminating wait doesn't just fail its own check — **it hands a false accusation to the next one**, naming a guard that was never broken. That is worse than a flake, and had I trusted the message instead of reading the blade I'd have gone hunting in entirely the wrong place.

I also patched only the two transitions I'd seen fail rather than auditing all three, which is why there was a round 4 at all.

**Every page transition in this function now waits on something true only at the destination.** That is the rule, not a patch.

## Why the failure now names its symptom

Earlier rounds failed at opaque `locator.waitFor: Timeout 10000ms exceeded`. That was partly because the check waited on the breadcrumb itself — the thing under test — so a breadcrumb defect surfaced as a navigation timeout rather than at the assertion written to catch it. Removing that dependency is what moved the mutated failure to `got 1`, which names the defect outright.

#141 closed unmerged.

---

PR #143, opened 2026-09-18
item/directory-access-ui: grant and revoke from the panel, and close the escalation it would have exposed (#14)

### LINKED ISSUE #14: Grant and revoke directory access from the UI
### Each `doneWhen` clause

- **A non-manager sees no control** — the Access section is gated by `@can('manageAccess', $selectedDirectory)`; tests cover a View-level holder and an Edit-level holder seeing neither `directory-access-panel` nor `grant-access-form`.
- **…and is refused server-side** — `grantAccess()` and `revokeAccess()` each authorise through the policy independently of what rendered, asserted with `assertForbidden()`; the actions authorise nothing (CLAUDE.md).
- **A grant is visible to the grantee in the same request cycle** — the test warms `DirectoryAccess`'s per-request memoisation to a cached *denial* before granting (its `resolve()` caches denials, not only hits), then re-asks the same singleton after. It fails if `DirectoryGrant::saved($flushAccess)` were unwired.
- **Revoke is covered** — happy path, cross-directory refusal, and a server-side refusal for a caller without manage.
- **Writes go through the `DirectoryGrant` model, never a mass query** — `->save()` and `->delete()` on instances. This is a seam rule, not style: a mass query bypasses the model events that flush `DirectoryAccess`, leaving stale access decisions for the rest of the request.

### The escalation this item exposed, and closed

`DirectoryPolicy::manageAccess()` required only `AccessLevel::Manage`, with no Spatie permission — the one sibling of spec §5's single capability ("grant/revoke access, move or delete the directory itself") skipping layer 1, and the most dangerous to leave open, since granting access is how every other restriction gets handed on.

Reachable by every user, not a corner case: `MEMBER_PERMISSIONS` has no `directories.manage`, and spec §5 says every user holds manage on their own home directory by ordinary grant. Latent only because nothing called the ability from the UI — **this item is what would have made it live**, so it was fixed here rather than filed.

The implementing worker found it and refused to resolve it inside the item, flagging instead. That was correct and was the most valuable part of its report.

Consequence, filed as #144: sharing is an administrator's act in v1. If self-service sharing is wanted, the answer is a permission of its own, not a layer removed.

### Smoke

`checkGrantAndRevokeDirectoryAccess()`, mutation-proven both directions — correct build green, `RevokeDirectoryAccess::handle()` neutered red at `revokeAccess() did not remove the DirectoryGrant row -- raw output: COUNT:1`, with the grant half still printing `OK` first so the failure is attributable to the one changed line.

abdullah-life:
Item 28. Grant and revoke directory access from the directory detail panel — plus a policy fix that had to land **with** it rather than after.

## The escalation this item would have exposed

`DirectoryPolicy::manageAccess()` asked only for `AccessLevel::Manage` on the directory. `move()`, `delete()` and `restore()` all additionally require `directories.manage`, and spec §5 groups them as one capability:

> **manage** — grant/revoke access, move or delete the directory itself.

So grant/revoke was the single sibling skipping layer 1 — and the most dangerous one to leave open, because granting access is how every other restriction gets handed to someone else. CLAUDE.md states the two-layer rule as binding.

It was reachable by **every user**, not a corner case. `MEMBER_PERMISSIONS` carries no `directories.manage`, and spec §5 also says every user holds manage on their own home directory through an ordinary grant. So any member could hand anyone any level on their own subtree, while being refused the move and delete the same sentence groups with it.

Nothing called the ability from the UI, so it was latent. **This item is exactly what makes it live**, which is why it is fixed here rather than filed — shipping the UI first would mean knowingly shipping the escalation.

**Nothing is taken away.** With no UI, nobody could grant through it. What this settles is that sharing is an administrator's act in v1. Whether an ordinary member should be able to share their *own* home directory is a real product question, wants a permission of its own rather than the removal of a layer, and is filed separately as #144.

The implementing worker found this and correctly refused to resolve it inside the item, flagging it instead. That was the right call and the most valuable part of its report.

## The pre-existing test is rewritten, not weakened

It asserted that a plain member with a Manage grant **passes** `manageAccess()` — documenting the ability as shipped rather than the rule. It now asserts each layer independently: grant alone refused, permission alone refused, only the pair passes. Neither layer can be dropped silently now.

## Two traps hit while writing it

**The permission-alone fixture must not be an admin.** `DirectoryAccess::resolve()` short-circuits `directories.view-all` to `AccessLevel::Manage` on *every* directory, so an admin would pass the level check by bypass and the assertion would prove nothing. It uses a member with `directories.manage` granted directly, as the rest of that file already does.

**The mutation anchor sits inside the `return` expression.** The clause is textually identical to `restore()`'s and `delete()`'s, and `mutation-check.php` matches by substring, so a bare anchor hit three sites. Anchoring *above* the return would leave a statement with no return, so the mutation would "fail" as a `TypeError` — testing that PHP has types, not that the test catches a missing layer. Verified by applying the `remove` string and running `php -l` on the result: valid, with the access check alone remaining.

All 33 `mutations.json` entries verified to match exactly once.

## The item itself

- `GrantDirectoryAccess` / `RevokeDirectoryAccess` actions, writing through **model instances** — never a mass query, because `DirectoryGrant::saved`/`deleted` flush `DirectoryAccess`'s per-request memoisation and a mass query bypasses them, leaving stale access for the rest of the request.
- `Browser::grantAccess()` / `revokeAccess()`, each authorising through the policy; actions authorise nothing (CLAUDE.md).
- A shared `directoryGrantsQuery()` so the revoke lookup and the rendered list cannot disagree — introduced by the worker to resolve a mutation-anchor collision, and a genuine single-source-of-truth improvement.
- Same-request-cycle test warms the memoisation to a cached denial *before* granting, then asserts visibility after, on the same singleton — so it fails if the flush wiring were removed.

Smoke: `checkGrantAndRevokeDirectoryAccess`. Predicted load-bearing assertion is the grant-row count reading 1 after Grant and 0 after Revoke; **prediction, not a claim** — mutation to follow.

Closes #14


abdullah-life:
## Mutation record

#145 makes `RevokeDirectoryAccess::handle()` a no-op — one line, nothing else touched.

**Correct build** — all 14 checks green including `image`, [job 105460286191](https://github.com/turbophp/doccum/actions/runs/35299949297/job/105460286191).

**Mutated build** — `image` fails, [job 105460305846](https://github.com/turbophp/doccum/actions/runs/35299956635/job/105460305846):

```
[setup] granting access through the panel wrote exactly one DirectoryGrant row -- OK
##[error][setup] revoking access through the panel did not remove the DirectoryGrant row
         -- raw output: COUNT:1
##[error]container-smoke (setup) failed: revokeAccess() did not remove the DirectoryGrant row
```

Both halves of what makes this evidence rather than noise:

- **The grant line prints first**, so the failure is attributable to the one changed line rather than merely co-occurring with it.
- **The failure names its symptom** with the raw count, instead of an opaque locator timeout.

## Getting there took correcting my own rule violation

The first mutation round reddened the job — but at `locator.waitFor: Timeout 10000ms exceeded ... 25 × locator resolved to visible <li data-test="directory-grant-row">`. The check had `Promise.all`'d the Revoke click with a wait for the grant row to become **detached**, which is precisely what revoking does. The wait duplicated the assertion and ran first.

That is exactly what `decision/0033` forbids, written by me one item earlier: *never wait on the thing under test, or a defect in it surfaces as an opaque timeout instead of at your assertion.* Knowing the rule wasn't sufficient — it took applying it to a second check before I started looking for the pattern rather than recognising it after the fact.

Now database-first, DOM-second, matching `checkBulkTrashLeavesUnselectedFilesAlone`'s existing docblock. The DOM check still runs afterwards and proves the panel re-rendered, which the database cannot — it is simply no longer what fails first.

## One choice that did pay off immediately

Mutating **revoke** rather than **grant** was deliberate: breaking the grant would have left no row for the revoke step to act on, so the check would have died on a locator before reaching any database assertion. That is how item 27's first two mutation rounds were wasted. Here the grant half survived and printed, which is what made the failure attributable.

#145 closed unmerged.

---

PR #147, opened 2026-09-18
item/trash-view: a Trash page whose listing is scoped in the query (#15)

### LINKED ISSUE #15: A Trash page with restore and purge
**Ledger id:** `item/trash-view`
**Size:** S
**Depends on:** `item/content-actions`, `item/topbar-shell`, `item/trash-purge-manual`

### Done when
Another user's trashed items never appear, asserted at query level; restore and purge are wired to the actions.

---

### LINKED ISSUE #86: Force-deleting a directory row would orphan every object beneath it
## The hazard

`database/migrations/…create_files_table.php:16`:

```php
$table->foreignId('directory_id')->constrained('directories')->cascadeOnDelete();
```

So `forceDelete()` on a `Directory` removes every descendant `File` **row** in SQL, directly. That
path never runs PHP per file, so it never reaches `Services\DocumentStorage` — every object those
files owned is left in the store with nothing referencing it.

`Directory::booted()`'s `forceDeleted` hook only clears the directory's own morph-attached
`properties`. It does not, and cannot, clean up descendants that the database already removed.

## Not currently reachable

Nothing in the application force-deletes a `Directory`. `PeriodPurger` force-deletes **files**, one
at a time, deleting their objects through `DocumentStorage` first — which is exactly the ordering
that makes it safe, and exactly what a cascade skips.

So this is latent, not a live bug. Filing it because the next thing to touch this area is likely to
trip it.

## Why it matters soon

`item/trash-purge-manual` (#11) deliberately scoped itself to **file** purge for this reason. The
item that wires purge into the UI, `item/trash-view` (#22), is where someone will reasonably reach
for "purge this directory" — and the obvious implementation, force-deleting the directory row, is
the unsafe one. It will look like it worked: rows vanish, the UI updates, and the bytes stay
forever.

## What a safe directory purge needs

Walk every descendant file first and purge each through the file path (objects before rows), then
remove the directory rows. Worth considering whether the FK should stop being `cascadeOnDelete` so
the database refuses the unsafe shortcut instead of silently performing it — a `restrict` would turn
this from a silent leak into a loud error, which is the `decision/0010` posture of stating the
semantic rather than inheriting the driver's convenience.

Surfaced while reviewing #11, where the constraint was verified against the migration rather than
assumed.

abdullah-life:
Item 29. A Trash page listing trashed files and directories the viewer may see, with Restore and Purge wired to the existing Actions.

**Purely additive: 748 insertions, zero deletions.** No existing Policy, Action, route or test changed.

## The central clause, enforced in the query

`Index::render()` narrows both listings with `whereIn` on ids resolved through `DirectoryAccess` — never a Blade `@if`, which would mean the view had already been handed rows it must hide:

- `viewableTrashedDirectoryIds()` is new and is the only place deciding which *trashed* directories a viewer may know about. It resolves through the same `can()` check `restore()`/`purge()` already use, so the listing can never surface something the Policy would refuse.
- Files are checked against the **union** of live and trashed reach, because a trashed file's directory may be live (trashed independently) or itself trashed (cascaded alongside it). `viewableDirectoryIds()` alone would miss the cascade case.
- Empty reach fails closed.

Asserted at query level rather than by page text — `viewData('files')->pluck('id')` compared exactly, not `assertDontSee`, which item 27 showed to be unreliable once a name can appear elsewhere on the page.

## Directory purge is deliberately absent, and that is #86

`PurgeFile`'s own docblock says it is *"Scoped to a single file on purpose, not a directory. A directory purge would need its own action: files.directory_id carries an ON DELETE [cascade] … Reclaiming a whole subtree's storage safely needs to visit every [file]"*. No `PurgeDirectory` exists, and nothing in `app/` force-deletes a directory.

That is exactly #86 — *force-deleting a directory row would orphan every object beneath it*. Building directory purge here would have introduced the caller that issue warns about. Directories get Restore only.

**A miss of mine, stated rather than smoothed over:** `decision/0031` resolved #86 by folding it into this item's `doneWhen`, and I never amended the text. The implementation independently respected the constraint, but it was luck of good judgement by the worker rather than the ledger doing its job. The `doneWhen` gains that clause in the completion ledger PR — recording a constraint the build already honoured, not relaxing one to fit.

## Unverified

`Livewire::test()->viewData()` appears **nowhere else in this suite**. It is documented, but this is the same shape of risk as `assertNotFound()` last item — which existed yet did not behave as assumed. CI settles it in one round; reasoning further without `vendor/` is what this loop keeps being punished for.

The worker also caught and fixed a real bug in its own test while hand-tracing (an assertion that would have passed regardless of the guard, because the fixture was never trashed) and flagged it unprompted.

Smoke: `checkTrashViewRestoreAndPurge()`, database evidence before DOM, no wait on the rows it acts upon. Its docblock notes it cannot see a missing `authorize()` call — both actors are admin — so that gap is the Pest tests' job, which is the right division and worth it having said so. Predicted load-bearing assertions are `RESTORE_LIVE:yes` / `PURGED_GONE:yes`; **prediction, not claim** — mutation to follow.

All 39 `mutations.json` entries verified to match exactly once.

Closes #15

---


abdullah-life:
## The mutation did not reach its assertion — and found a different defect

#148 failed, but **not at `PURGED_GONE`**. It never got there. It died in this check's *setup*:

```
[setup] restoring and purging trashed files through the new /trash page (issue #15)
[setup] DoccumSmokeTrashViewRestore.txt stored, confirmed by a files row
##[error]DoccumSmokeTrashViewPurge.txt was uploaded and left no files row -- issue #106 has regressed
```

So the mutation result is **void**, not negative: with the run aborted before either click, it says nothing about whether `PURGED_GONE:yes` is load-bearing. Per CLAUDE.md I can't cite this PR's smoke evidence until it does, so #148 has been rebuilt on the fix below and re-run rather than written off.

### What actually happened

`checkTrashViewRestoreAndPurge()` is the first check that uploads **two** files into one page. The probe's reads for the second upload:

```
POST .../update        req 53.4297  resp 53.4644   (_startUpload)
POST .../upload-file   req 53.4671                 (the bytes)
POST .../update        req 53.5004  resp 53.5601   <-- the submit commit
                       resp 53.5684                (upload-file answers)
POST .../update        req 53.5741  resp 53.6044   (_finishUpload, 30ms)
```

The submit commit went out **67ms before the bytes were acknowledged**, so `store()` ran against a property Livewire had not populated yet. That is issue #106's discard exactly — reached through the harness, not the product. The 30ms tail commit is `_finishUpload`, far too fast to be a `store()` that wrote to MinIO; compare the first upload in the same run, whose `store()` took 982ms.

### Why `clickOnceUploadSettles()` did not prevent it

It reads the upload's progress off the submit button's `disabled` attribute. A poll that latches onto the first enable it sees cannot distinguish *"this upload has finished"* from *"some upload has finished"* — neither the tail of the previous upload nor a gap between `_startUpload`'s response and the upload POST is distinguishable from the real edge. That is why this only surfaced on the first check to upload twice in a row.

### The fix (c7fe0cb)

`setFileAndWaitForUpload()` waits on the **upload's own request lifecycle** — the `upload-file` response, with the listener armed *before* `setInputFiles()` because a file input posts on change — then lets `_finishUpload` settle. `clickOnceUploadSettles()` stays as the last-mile guard for a button wedged shut by a stuck loading state, which a request wait cannot see; in a healthy run it now falls straight through.

Two things deliberately unchanged:

- **The product guard.** `wire:loading.attr="disabled" wire:target="upload"` is untouched. This is a harness repair, and the honest label matters: `decision/0030` records the opposite mistake — #106 §2 hypothesised the real product race and filed it against the smoke, so the harness got patched while the product kept the defect. This one genuinely is the harness: a person sees a disabled button and clicks when it looks clickable.
- **`checkUploadButtonIsDisabledWhileTheFileIsStillUploading()` still sets its file by hand.** It needs the raw input — a helper that waits the upload out would wait out the stall it installs on purpose.

No retry was added. #129 removed the retry here precisely because a retry converts a timing defect into a green run, which is how #106 survived three rounds.

---

abdullah-life:
## Mutation re-run: `PURGED_GONE` is load-bearing — proven

#148 rebuilt on c7fe0cb (`PurgeFile::handle()` still a no-op, nothing else changed) reached its assertion this time, and failed exactly where the PR body predicted:

```
[setup] DoccumSmokeTrashViewRestore.txt stored, confirmed by a files row
[setup] DoccumSmokeTrashViewPurge.txt stored, confirmed by a files row
[setup] DoccumSmokeTrashViewRestore.txt is live again -- Restore on /trash works
##[error][setup] DoccumSmokeTrashViewPurge.txt still exists after Purge on /trash
         -- raw output: RESTORE_LIVE:yes PURGED_GONE:no
```

Both directions are now on record: `image` **green** on #147 (c7fe0cb, all 14 checks), `image` **red** on #148 at `PURGED_GONE:no`, with the restore half passing *and printing* first so the failure is attributable to the one line that changed. `decision/0034`'s rule — mutate the step the check can still reach past — did what it was written to do.

Both uploads also stored on both branches, which is the smoke fix (c7fe0cb) working in both directions rather than only in the green one.

### What is NOT proven, said rather than implied

**`RESTORE_LIVE:yes` remains a prediction.** It passed in both directions here — necessarily so, since neutering purge is what let the restore half print at all. Proving it would need its own mutation against `RestoreFile`. CLAUDE.md requires *at least one* shown-to-fail assertion per v1 UI item and `PURGED_GONE` is that one; I am not stretching it to cover its neighbour.

### One more mutation before this merges: #149

#147 moves `uploadAndProveStored()` off the button's `disabled` attribute, so that helper is no longer sensitive to issue #106's product guard. The whole guard now rests on one assertion in `checkUploadButtonIsDisabledWhileTheFileIsStillUploading()` — and `decision/0030` called that assertion "load-bearing" on reasoning, with no mutation run behind it. That was tolerable when a second check was also sensitive; it is not tolerable now that it is the only one.

turbophp/doccum#149 removes `wire:loading.attr="disabled" wire:target="upload"` from the Upload button and nothing else. This PR does not merge until that comes back red.

---

PR #148, opened 2026-09-18
MUTATION — DO NOT MERGE: purge does not delete the file

abdullah-life:
Not for merge. #147's smoke mutation. Opened as a PR because `tests.yml` runs the `image` job on `pull_request` and pushes to `main` only, so a bare branch push produces no image job and its silence reads like success (`run/0014`).

`PurgeFile::handle()` returns immediately. Nothing else is touched.

## Why purge and not restore

`decision/0034`: mutate the step the check can still reach past. The check reads `RESTORE_LIVE` and `PURGED_GONE` from **one** `tinker` call and asserts restore first — so with purge neutered, the restore half still passes *and prints*, and the failure arrives at the second named assertion with the first half's success as context.

Breaking restore instead would still reach both reads, but the earlier assertion would fail and the later one — the one the mutation is meant to exercise — would never be evaluated.

**Expected: `image` FAILS** at `PURGED_GONE`, *after* printing `… is live again -- Restore on /trash works`.

If it passes, that assertion proves nothing and #147's smoke evidence is worthless — which is the whole reason this exists. The implementing worker named `RESTORE_LIVE:yes` / `PURGED_GONE:yes` as its predicted load-bearing assertions and explicitly declined to claim them proven.

Closed and left unmerged once the result is recorded on #147.

---


abdullah-life:
Closed unmerged, as intended. Result recorded on turbophp/doccum#147.

The first run of this branch was **void, not negative** — it died in the check's setup (the second upload stored nothing) before either click, so it said nothing about `PURGED_GONE`. Rebuilt on c7fe0cb, it produced the verdict it exists for:

```
[setup] DoccumSmokeTrashViewRestore.txt is live again -- Restore on /trash works
##[error][setup] DoccumSmokeTrashViewPurge.txt still exists after Purge on /trash
         -- raw output: RESTORE_LIVE:yes PURGED_GONE:no
```

---

PR #149, opened 2026-09-18
MUTATION -- DO NOT MERGE: the Upload button carries no in-flight guard

abdullah-life:
Not for merge. Opened as a PR because `tests.yml` runs the `image` job on `pull_request` and pushes to `main` only, so a bare branch push produces no image job and its silence reads like success (`run/0014`).

`wire:loading.attr="disabled" wire:target="upload"` is removed from the Upload submit button in `resources/views/livewire/files/browser.blade.php`. Nothing else is touched.

## Why this exists now

#147 moves `uploadAndProveStored()` off the button's `disabled` attribute and onto the upload's own request lifecycle, because timing the click off that attribute was silently clicking mid-upload (see #147's comment). That is the right repair, and it has a cost: **that helper is no longer sensitive to the product guard.** With the bytes acknowledged before the click, an undisabled button stores just fine.

So the whole of issue #106's guard now rests on one assertion, in `checkUploadButtonIsDisabledWhileTheFileIsStillUploading()`, which stalls `**/upload-file*` on purpose and asserts the button disabled inside that window.

`decision/0030` already calls that assertion "load-bearing". Reading back, that was a **prediction stated as a finding** — I can find no mutation run behind it. CLAUDE.md is explicit that this is exactly the reasoning that fails: the topbar check's load-bearing half was argued one way and mutation proved the reverse. It was tolerable while a second check was also sensitive to the guard. It is not tolerable now that this is the only one.

**Expected: `image` FAILS** with:

```
the Upload button stayed ENABLED while the upload endpoint was stalled --
issue #106 is reachable again: ...
```

and it must fail **after** printing `the Upload button starts enabled (control, not evidence)` — that control line proves the check reached its subject rather than dying on the way.

If it passes, the guard has no proof at all and #147 does not merge until it has one.

Closed and left unmerged once the result is recorded on #147.

---


abdullah-life:
Closed unmerged, as intended. The guard's assertion is load-bearing — now on a run rather than on reasoning:

```
[setup] stalling the upload endpoint to check the Upload button is disabled in flight (issue #106)
[setup] the Upload button starts enabled (control, not evidence)
##[error]container-smoke (setup) failed: the Upload button stayed ENABLED while the
         upload endpoint was stalled -- issue #106 is reachable again: either
         wire:loading.attr="disabled" wire:target="upload" is gone from
         browser.blade.php, or Flux is no longer forwarding those attributes to the
         real <button>
```

It failed **after** printing its control line, so it reached its subject rather than dying on the way, and it failed with its own message rather than a locator timeout — the two things that make the verdict attributable to the one attribute that was removed.

That also settles the second half of the question this PR was opened on: **Flux does forward `wire:loading.attr` to the real `<button>`.** The check cannot distinguish the two causes in its error text, but the mutation can: removing the attribute is what turned it red, so the attribute was reaching the DOM before.

Recorded on turbophp/doccum#147 and in `decision/0035`. `decision/0030`'s claim stands, with the provenance it should have had.

---

PR #152, opened 2026-09-18
item/download-missing-object: a missing object is a stated answer, not a crash (#134)

### LINKED ISSUE #134: Download answers HTTP 500 when the file row exists but its object is gone
Observed, not theorised. #132 deliberately empties `/data/objects` between the two smoke containers, leaving SQLite intact, and the replacement container then serves a file listing whose rows all resolve and whose bytes do not exist:

```
[verify] the uploaded file is still listed -- the DATABASE persisted
[verify] following the Download link for smoke-...-1.txt: http://127.0.0.1:8080/files/1/download
[verify] Download for smoke-...-1.txt answered HTTP 500
```

That is a real operator state, not only a mutation artifact: a restore that brought back the database but not the objects, a bucket removed out of band, a partially restored volume. The container itself stays up and every other page works.

## What is worth deciding

A 500 is defensible — a missing object genuinely is an exceptional server state, and doccum should not pretend the document is absent when the row says it exists. But 500 is what an unhandled exception also produces, and from the outside an operator cannot tell "your object storage is missing" from "doccum crashed". The two need different responses from them.

I have not read what actually produces the 500 (an unhandled exception out of the streaming path, or a deliberate abort), so the first step is establishing that rather than assuming it. `Services\DocumentStorage` is the only code that resolves a disk, so whatever the answer, that is the seam.

## Not scope creep on #131

#131 asserts that the bytes come back and says nothing about what should happen when they cannot. Filing rather than fixing, so the new LOOP.md step-1 sweep disposes of it deliberately — this is a good candidate for folding into an existing item's `doneWhen` rather than becoming an item of its own.

Found by the mutation for #131.

abdullah-life:
Item 30. A `file_versions` row whose object is gone now answers a deliberate 502 with the object key in the body and in a log line, instead of whatever status happened to fall out.

## The fix is structural, not a translated exception

`readStream()` was called **inside** the `streamDownload()` callback, which Symfony runs during `sendContent()` — after the status line has gone out. So the 500 observed through FrankenPHP was an accident of when the exception surfaced relative to the header flush; a different server could send `200` with an empty body.

Catching and re-throwing would not have fixed that. The stream is now opened **before** the response is constructed, which turns "what status does a missing object get" back into a decision. The open handle is captured into the callback and `fclose()`d there as before, so nothing leaks on the success path.

`DocumentStorage::readStream()` throws the new typed `ObjectMissingFromStorage` instead of a bare `RuntimeException`, so callers can distinguish "the row exists, the bytes don't" from any other storage failure.

## Why 502, stated in the code

Object storage is genuinely upstream of PHP for the MinIO/S3 case doccum ships with, so "the upstream returned nothing" is honest there. For the embedded local disk it's a stretch — no real network hop — but **a stated stretch beats an unhandled 500**, and 404 would say "this file doesn't exist", which is false: the row and every other version may be intact.

## Why this earns v1

`/data` is the only persistent volume and the backup story is "copy `/data`". A restore that brings the database but not `objects/` is the likeliest operator mistake in the product, and today it presents as "doccum crashed". `item/self-hosting-docs` has to be able to say what happens in that state.

## Changed after the worker handed it back

The tests asserted the log line with `Log::shouldReceive('error')`, which installs a **strict** Mockery mock on the facade — any *other* `Log::` call in the request (a deprecation, a framework notice) raises `BadMethodCallException` and reddens the test for a reason unrelated to what it asserts. Switched both to `Log::spy()` + `Log::shouldHaveReceived()`, which permits every call and asks afterwards about the one that matters. Without `vendor/` that failure mode would have surfaced only in CI, looking like the fix failing rather than the harness misfiring.

The worker correctly reported that **no `Log::` assertion exists anywhere in this suite** — my brief claimed there was precedent to copy, wrongly. So this sets the convention, and the comment in both tests says so.

## Unverified, flagged rather than smoothed over

`Storage::fake('documents')` is called with no `$config`, and the real `documents` disk sets `'throw' => true`. The tests depend on the fake **not** carrying that flag forward, so `readStream()` returns `null` rather than throwing Flysystem's `UnableToReadFile`. That is my reading of `Storage::fake()`, which builds a local driver from the passed config only — but `vendor/` is absent here, so neither the worker nor I could read the source or execute it. **CI settles it in one round.** If it goes the other way the fix is a one-line `Storage::fake('documents', ['throw' => false])`, not a change to the product code.

## Mutation to follow

Predicted load-bearing: `assertStatus(502)` paired with the `shouldHaveReceived('error')` expectation — reverting the fix should miss the log call entirely and produce something other than 502. **Prediction, not claim.** The mutation runs before this merges, per `decision/0035` — an item whose evidence is asserted rather than demonstrated is exactly what that decision was written about.

## Reported, not fixed

`downloadToTemp()` has the identical "null stream → throw" shape and still raises a bare `RuntimeException`. It feeds text extraction, not a download route, so it is out of this item's scope — filing separately rather than widening the PR.

Closes #134

---


abdullah-life:
## CI found a bigger bug than the one this PR was opened for

All six test jobs answered **500 where the tests expect 502**, deterministically. Not a flake, and not the tests being wrong.

`config/filesystems.php:71` sets `'throw' => true` on the `documents` disk. Laravel's `FilesystemAdapter::readStream()` **rethrows** Flysystem's `UnableToReadFile` when that flag is set, and returns `null` only when it isn't.

So:

- **`if ($stream === null)` has been dead code in the shipped product** for as long as that flag has been set.
- A missing object has always surfaced as `UnableToReadFile`, raised from inside the `streamDownload()` callback — which is exactly the 500 #134 reported.
- The typed `ObjectMissingFromStorage` added in `25c9d16` **could never have been thrown by a real install.** Converting only the null branch would have turned these tests green against a code path nobody takes — a fix that reports success and changes nothing, which is the failure mode CLAUDE.md is about when it says a green suite is necessary and not sufficient.

`70ff6cc` catches `UnableToReadFile` at the seam and converts it, keeping the null branch for a disk configured without the flag. Which signal you get depends on disk config rather than on anything the caller controls, so `DocumentStorage` owes its callers one answer for both.

**The controllers are unchanged and were always right** — opening the stream before constructing the response is the structural fix, and it stands. Only the *type* of exception crossing that boundary was misidentified: by the issue, by the brief I wrote, and by me when I reviewed the diff and signed off on it.

### On the thing I flagged as unverified

The PR body said the risk was `Storage::fake()` inheriting `'throw' => true`, and that if it went the other way the fix would be a one-line `Storage::fake('documents', ['throw' => false])`. **That would have been the wrong fix** — it would have made the tests pass by configuring the fake to behave *unlike* the real disk, hiding the production defect completely. Flagging the uncertainty was right; the remedy I proposed for it was not.

---

abdullah-life:
## Mutation verdict: negative, with the control passing

turbophp/doccum#154 moved `readStream()` back inside the `streamDownload()` callback in `FileDownloadController` only:

```
FAILED  Tests\Feature\FileDownloadTest > it answers 502 with a logged obj…
  Expected response status code [502] but received 200.

Tests: 1 failed, 1 skipped, 695 passed (1431 assertions)
```

One failure, and it is the right one. `FileVersionDownloadTest` — the untouched control — passed, so the harness reached its subject rather than breaking generally. That is the distinction `decision/0035` was written about.

### The status it received is the finding

**200, not 500.**

The PR body offered this as reasoning: *"the 500 observed through FrankenPHP is an accident of when the exception surfaces relative to the header flush; a different server could send 200 with an empty body."* That was speculation. It is now measured — under PHPUnit's server, the same code answers **200 with a broken body**.

So the item's central claim is demonstrated rather than argued: a missing object discovered inside the callback cannot choose a status, because the status has already gone out. The observed code genuinely is an artefact of the server, and on at least one of them the failure is **invisible to any client checking the status code**.

### Both halves are load-bearing, independently

| | | |
|---|---|---|
| Structural fix without seam conversion | round one of this PR | **500** — not sufficient |
| Seam conversion without structural fix | #154 | **200** — not sufficient |
| Both | `70ff6cc` | **502**, 14/14 green |

Neither half would have fixed #134 alone, and either one alone would have looked like progress. Worth stating because the item was scoped as one change and turned out to be two.

`RESTORE_LIVE`-style caveat: nothing here proves the `Log::shouldHaveReceived('error')` expectation is independently load-bearing — it rode along with the status assertion in the same failure. The status assertion is the one shown to fail.

#154 is closed unmerged. This merges next.

---

PR #154, opened 2026-09-18
MUTATION -- DO NOT MERGE: the stream is opened inside the callback again

abdullah-life:
Not for merge. The mutation for #152, opened as a PR so the full `test` matrix runs.

`readStream()` moves back inside the `streamDownload()` callback in **`FileDownloadController` only**. The typed exception, the catch, the log line and the 502 go with it — once the status line is already decided there is nowhere left to put them, which is exactly the thing under test.

## What this isolates, and why it is worth a round

CI already proved these are two different fixes rather than one:

1. **Convert the exception at the seam** — `DocumentStorage` turning Flysystem's `UnableToReadFile` into `ObjectMissingFromStorage`. Untouched here.
2. **Open the stream before building the response** — the structural half. Reverted here.

Round one of #152 had (2) without (1) and every test job answered 500. So (2) alone is demonstrably not sufficient. This asks the converse: **is (2) necessary, or would (1) alone have been enough?** If the tests pass with the stream opened inside the callback, then the item's central claim — that the status must be decided before any header is sent — is not something these tests can see, and the `doneWhen` is asserted rather than demonstrated.

## The control

`FileVersionDownloadController` is deliberately untouched.

- **Expected:** `FileDownloadTest`'s 502 case **FAILS**; `FileVersionDownloadTest`'s **PASSES**.
- If **both** fail, the harness broke rather than the guard, and this run is void — the distinction `decision/0035` was written about, after a mutation went red in the right file for the wrong reason and nearly got recorded as proof.
- If **neither** fails, #152's tests do not test what the item claims and it does not merge in this shape.

Predicted load-bearing: `assertStatus(502)` together with `Log::shouldHaveReceived('error')`. Prediction, not claim — that is what this run is for.

Closed and left unmerged once the result is recorded on #152.

---


abdullah-life:
Closed unmerged, as intended. Verdict **negative** — the good outcome — recorded on turbophp/doccum#152.

```
FAILED  Tests\Feature\FileDownloadTest > it answers 502 with a logged obj…
  Expected response status code [502] but received 200.

Tests: 1 failed, 1 skipped, 695 passed (1431 assertions)
```

One failure, in the mutated controller's test. `FileVersionDownloadTest` — the untouched control — passed, so the harness reached its subject rather than breaking generally.

The **200** is the part worth keeping. The PR body predicted the mutation would fail; it did not predict *this*, and it is stronger evidence than a 500 would have been. With the stream opened inside the callback, the status line is already committed as success before the missing object is discovered, so on this server the failure is invisible to any client that checks only the status code. #134's original report of a 500 through FrankenPHP really was an artefact of that server, exactly as the item argued and could not previously show.

---

Issue #150, opened 2026-09-18
Mutation runs should be ledger nodes with verdicts, so a prediction cannot pass as a finding


abdullah-life:
## The gap

Hand-run image mutations are recorded as `PullRequest` nodes with `state: closed` and a name beginning `MUTATION -- DO NOT MERGE` (PRs #118, #122, #130, #132, #141, #145, #148, #149). That convention lives entirely in a string. From the vocabulary alone, #145 is indistinguishable from an abandoned branch — and **nothing records the verdict**.

Three things happened this hour that the ledger cannot express:

1. **#148's first run was void, not negative.** It went red, on the mutation branch, in the check the mutation targets — which looks exactly like the verdict it was opened to collect. It wasn't: it died in the check's *setup*, before either click. Reading the conclusion instead of the log would have recorded a proof that did not exist. The ledger says `state: closed`.
2. **`decision/0030` recorded a prediction as a finding.** It called an assertion "load-bearing" with no run behind it. That went unnoticed for four items, and was caught only because a later fix made that assertion the *only* proof of issue #106's guard — at which point #149 had to be run to earn the claim retroactively.
3. **`RESTORE_LIVE:yes` is a prediction** and is correctly labelled as one in prose, in a PR comment and a decision node. Prose is where the last one hid too.

## Shape

A top-level `mutations` list on the Ledger, sibling to `pullRequests` and `decisions`; node type `Mutation`. Deliberately **not** nested inside `PullRequest`: a void run and its rebuild are two mutation *runs* that can share one PR (a force-push), so mutation ≠ PR.

```jsonc
{
  "@id": "mutation/0009",
  "@type": "Mutation",
  "implements": ["item/trash-view"],
  "pullRequest": "https://github.com/turbophp/doccum/pull/148",
  "run": "run/0018",
  "headSha": "22d0e37…",
  "mutant": "PurgeFile::handle() returns immediately, so the row survives forceDelete",
  "check": "PURGED_GONE:yes in checkTrashViewRestoreAndPurge()",
  "verdict": "negative",
  "url": "https://github.com/turbophp/doccum/actions/runs/…",
  "supersedes": "mutation/0008",
  "rationale": "…"
}
```

`negative` is the **good** outcome — the check went red. The word should stay uncomfortable so nobody misreads `positive` as "passed". `positive` means the guard is not load-bearing and must block the item. `void` means no evidence either way.

## Validator rules that make it bite

[...]

PR #160, opened 2026-09-18
item/home-dashboard: recents filtered by DirectoryAccess, and one link named Home (#16)

### LINKED ISSUE #16: Home dashboard: recents, storage by period, quick search

### LINKED ISSUE #140: Two links read "Home" on /files, with different destinations
Introduced by `item/files-three-pane` (#139), flagged there rather than fixed, to keep that PR to its item.

On any `/files` page there are now two links whose accessible name is exactly `Home`:

- `resources/views/layouts/app/topbar.blade.php` — the app home
- the Files sidebar's pinned link — the **viewer's own home directory**

Different destinations, identical accessible name, same page. For a screen-reader user listing links that is genuinely ambiguous; it is not merely a test-locator inconvenience.

## Why it is not urgent

A `data-test="sidebar-home"` hook exists on the sidebar entry, so Playwright can disambiguate by scope, and no current locator uses `getByRole('link', { name: 'Home' })`. Nothing is broken today.

## What to decide

Whether the sidebar entry should keep the label `Home` at all. Candidates: the directory's own name (which is the username, so it would then duplicate the tree's first entry), `My files`, or an `aria-label` distinguishing it while keeping the visible text. This is a wording and information-architecture call against spec §10, not a bug with one obvious fix — which is why it is filed rather than patched.

Related: the same PR had to rescope 13 existing Playwright locators because the sidebar made a previously unique directory name ambiguous. Accessible-name collisions on this page have already cost a round once.

abdullah-life:
Item 32, the first on **run/0019** — which matters, because rule 1 from #150 binds from run/0019 and this is a `spec:10` item. It cannot be recorded Completed without a Mutation node naming an assertion shown to fail.

## Recents, filtered in the query

`Home\Index` resolves `viewableDirectoryIds()` through `DirectoryAccess` and narrows with `whereIn` — never a Blade `@if`. Trashed files are excluded by `File`'s own `SoftDeletes` scope rather than an extra condition, which the worker confirmed by reading the model rather than assuming.

Zero-data is a real clause and is tested twice: component-level and over HTTP. A brand-new user with no reachable directory gets a working page.

## The two links named "Home" (#140, folded)

Spec §10 names **both** the topbar entry and the sidebar's pinned entry "Home", so the collision is the spec's. The sidebar entry keeps its visible text and gains `aria-label="Home directory"` — which *contains* the visible text, so WCAG label-in-name holds. Not the username (the tree's first entry already reads that), not "My files" (a term the spec never uses).

A plain `<a>` rather than `flux:link`, and the reasoning is worth keeping: Flux is only *known* to forward arbitrary attributes on `flux:button`. If it placed `aria-label` on a wrapping element, the **anchor's own** accessible name — computed from its own attributes and content — would be unchanged, and the fix would silently do nothing while looking correct.

## The clause neither of us could run, answered better than asked

The `doneWhen` asks for the accessible-name test "shown failing against main first". There is no `vendor/` here, so neither the worker nor I could demonstrate that.

Rather than claim it, I added a **`mutations.json` entry** — `sidebar-home-carries-a-distinct-accessible-name`. Removing the `aria-label` drops the anchor's accessible name back to "Home", both links read the same, and the test fails. `guards` re-proves that on **every CI run**, which is strictly stronger than a one-off demonstration and uses the mechanism this repo already has for it.

## Verification I did myself

All **41** `mutations.json` entries: every `remove` string matches its file exactly once, and every `expectFailing` names a test that exists. I now check the second against the test sources rather than by eye — my first attempt grepped for a test name and got a false negative on an escaped apostrophe (`user\'s`), which would have had me "fixing" a correct entry.

## Mutation still to run

Predicted load-bearing: `checkHomeDashboardShowsRecentUpload()`'s wait for `[data-test="home-recent-file-row"]` containing the uploaded filename. **Prediction, not claim** — the image mutation runs before this merges, and rule 1 now enforces that rather than trusting me to remember.

## Deliberately not built, reported as follow-ups

Spec §10's Home also names recent activity and storage-by-period. Neither is built:

- **Recent activity** — no audit log exists anywhere in the codebase. Follow-up: `item/audit-activity-log`.
- **Storage by period** — needs a bytes-per-period aggregate that doesn't exist; `doccum:purge-period` reports one period on demand. Follow-up: `item/home-storage-by-period`.

The item is size S. A dashboard that grows into an L is a split, not a widening.

## Noticed, not fixed

`routes/web.php` has no `declare(strict_types=1);`, contrary to CLAUDE.md. Pre-existing and untouched.

Closes #16
Closes #140

---


abdullah-life:
## Mutation verdict: negative — and it was not the only guard, as predicted

turbophp/doccum#162 left `Home\Index` untouched and stopped the view rendering rows. `image` failed inside the named check with every prior check green:

```
[setup] smoke-35313418734-1.txt stored, confirmed by a files row
[setup] checking Home lists the just-uploaded file under Recent files (issue #16)
```

The `test` matrix went red too, exactly as the mutation PR said it probably would: `HomeDashboardTest` asserts rendered output at lines 85 and 132, so it is independently sensitive to the same breakage.

**So the honest reading is that this smoke assertion is real but not the sole guard here.** I picked the mutation that answers the question rather than one only the smoke could catch — a mutation chosen so that nothing else can see it proves something about the mutation, not about the assertion. Recorded that way in the Mutation node.

## The accessible-name clause is now proven too

The same run's `guards` log carries the clause the `doneWhen` asked for and that neither the worker nor I could demonstrate locally:

```
guard present  -> "has exactly one link on /files with the accessible name "Home"" passes
guard deleted  -> "has exactly one link on /files with the accessible name "Home"" fails, as it must
```

"Shown failing against main first" — except it now re-proves itself on every CI run instead of once.

(`guards` also reported `home-recent-files-query-scoped-to-reach` as unprovable on that branch: my blade mutation breaks the `assertSee` inside the very test that entry depends on, so it "does not pass with the guard in place". Expected noise on a branch that never merges; that entry passes here.)

## One weakness fixed (`466ebdd`)

The mutation failed with a bare `locator.waitFor: Timeout 10000ms exceeded` — naming no assertion. It was attributable only because this check's console line happened to be the last thing printed, which is luck rather than evidence, and **rule 3 of the vocabulary I shipped two items ago says exactly that**. It now fails with `Home listed no recent-files row for <name>` plus container state, like every other check in the file.

The assertion is unchanged — same locator, same wait — so #162's verdict still describes what this code asserts. I am not re-running the mutation for a changed error string, and I am saying so rather than implying the recorded verdict covers this head byte-for-byte.

---

PR #165, opened 2026-09-18
item/search-filters: every filter inside the query, permission scope intact (#17)

### LINKED ISSUE #17: Type, mime, period and property filters on search
**Ledger id:** `item/search-filters`
**Size:** M
**Depends on:** `item/ci-green-baseline`

### Done when
Each filter is applied inside the query with the permission filter still active; the property filter uses the typed value columns.

---

abdullah-life:
Item 33. Type, mime, period and property filters, applied **inside the query** in both index backends, ANDed onto the permission scope rather than replacing it.

## The clause that matters

`Services\Search` exists so no caller can forget to scope by reach — its own docblock says a caller that forgets "leaks the whole corpus, and it would look like working code." Every filter here is chained onto the same builder / the same `WHERE` that already carries `viewableDirectoryIds()`. No filtering of a returned Collection: that would both break `limit` semantics and put the whole matching set in memory first.

Every property-filter test set includes a case where an **unreachable directory carries the same matching property value**. A filter test that only proves narrowing is half a test.

## Both backends, confirmed rather than assumed

`subject_type`, `mime`, `period_year`, `period_month`, `extension` were already applied identically in both. The new property filter goes through a shared `PropertyFilter` helper so the type→column mapping cannot drift, with the same `EXISTS` shape and the same correlation keys in each.

One documented divergence: only `LikeSearchIndex` on PostgreSQL adds the `md5(value_string)` predicate — that index exists only on PostgreSQL, and FTS5 never runs there. The `properties` migration's own comment says `item/search-filters` must add it or the index is dead weight.

## What I changed after the hand-back, and why it mattered

The worker flagged, rather than hid, that its FTS5 guard **failed for the wrong reason**. `fts5-property-filter-keeps-permission-scope` removed the SQL fragment `AND d.directory_id IN (...)` alone — leaving the directory ids in `$bindings`, desyncing the placeholder count, and making SQLite throw.

That guard would have "failed when removed" by proving **SQLite counts placeholders**, not that the predicate keeps unreachable documents out. `DirectoryPolicy`'s entry spells out the same trap for a policy clause: an anchor that leaves invalid code tests that PHP has types rather than that the test catches a missing layer. This was its search-side twin, reachable because the predicate and its bindings were built in two separate places, so no contiguous removal could take both.

They are now built together, both defaulting to empty above the block. Removing it leaves **valid SQL that simply stops restricting by directory** — a real leak. Simulated locally: the mutated file passes `php -l`, the predicate is gone from the query, and both variables survive as their empty defaults.

The `LikeSearchIndex` twin already demonstrated a real leak (removing a `whereIn()` leaves working code) and is untouched.

All **43** `mutations.json` entries re-verified: every `remove` matches its file exactly once.

## Possible CI finding, stated in advance

The restructure assigns `$permission`/`$permissionBindings` to empty and then immediately reassigns them. That is deliberate — it is what makes the block removable — but **phpstan may flag the redundant assignment**. If it does, that is mine to fix, and I will not "fix" it by making the guard unremovable again.

## Mutation to follow

Predicted load-bearing smoke assertion: the **excluded half** of `checkSearchFilterExcludesByPeriod()` — `waitFor({state:'detached'})` after filtering to a year the file cannot carry. The worker deliberately avoided asserting on an element already visible, which CLAUDE.md's topbar precedent shows proves nothing. **Prediction, not claim**; rule 1 will refuse the Completed record without it.

## Noticed, not fixed

- No `mutations.json` entry covers the pre-existing permission predicate *on its own* — the two here exercise it only via the property filter. Out of this item's scope; flagged for a sweep.
- The mime filter is free-text exact match rather than a picker of known values. Possible follow-up: `item/search-mime-picker`.

Closes #17

---


abdullah-life:
## Mutation verdict: negative, and the engine split showed up in the test matrix

turbophp/doccum#166 dropped `period_year` from the filter list `Fts5SearchIndex` iterates. Nothing else changed.

```
[setup] search found the uploaded document -- OK
[setup] checking the search page's period filter actually narrows results (issue #17)
##[error][setup] the period filter did not exclude smoke-35317181590-1.txt when Year was set to 2031,
         a year it cannot carry
```

Failed at the named assertion, with its own message, immediately after the check's console line, every prior check green. No anonymous locator timeout — that is item 32's rule-3 fix paying off one item later.

## The unplanned finding: per-engine coverage is real

The `test` matrix split exactly along the backend boundary:

| engine | index | result |
|---|---|---|
| sqlite 8.4 / 8.5 | `Fts5SearchIndex` | **failed** |
| pgsql 8.4 / 8.5 | `LikeSearchIndex` | passed |
| mysql 8.4 / 8.5 | `LikeSearchIndex` | passed |

I mutated only the FTS5 path and precisely the SQLite jobs went red. That is incidental evidence for the thing this item was most at risk of getting wrong — **a filter implemented in one backend and forgotten in the other really would surface as a per-engine failure**, rather than being masked by duplicated-but-vacuous assertions. The two `LikeSearchIndexTest` / `Fts5SearchIndexTest` files are exercising genuinely different code.

## What this does not prove

Only the **period** filter is mutation-proven. Mime, type and property have tests including out-of-reach cases, and the two property-filter permission guards are in `mutations.json` so `guards` re-proves them every run — but the smoke assertion shown to fail is the period one. Recorded that way.

#166 is closed unmerged. #165 is green 14/14 and merges next.

---

PR #168, opened 2026-09-18
item/login-by-username: accept a username at login, and enforce what makes that unambiguous (#75)

### LINKED ISSUE #75: Login accepts only email, but spec §10 says either username or email
## Gap

Spec §10 states that login accepts **either username or email**. Only email works.

Confirmed while implementing `item/email-folding` (#59): before that change there was no
`Fortify::authenticateUsing()` customisation anywhere in the codebase, no custom guard or
user provider, `config/auth.php` uses the stock Eloquent provider, and the login form has
only an `email` field.

`app/Actions/Fortify/AuthenticateUser.php` (added by #59) is now the only
`authenticateUsing` customisation, and it resolves by `email` alone — deliberately, and
documented as out of scope in its docblock, so that this remains a separate, argued change.

## What it needs

Not just a second `where`:

- **A collision policy.** `users.username` and `users.email` are separate columns with
  separate uniqueness. A submitted credential matching one user's username and a different
  user's email has to resolve deterministically, and the choice has to be stated rather
  than fall out of query order.
- **One folding rule per column.** `App\Support\EmailKey` folds email (case only, no NFC).
  Usernames are not necessarily the same rule, and per `decision/0010` doccum states the
  semantic rather than inheriting the driver's — so whichever rule usernames get must be
  explicit and persisted the same way, not left to a collation.
- **The login form label**, currently "Email address".
- **Tests across all six CI legs**, since this is exactly the cross-driver comparison class
  that produced #59.

abdullah-life:
Item 34. `AuthenticateUser` resolves an identifier that is either a username or an email; the login field stops being `type="email"`, which rejected a username in the browser before any of this was reached.

## The `doneWhen`'s collision clause, corrected rather than satisfied literally

It asks for "the collision rule decided and stated" for *a value that is a valid username **and** some other account's email*. I checked before briefing: **that collision is unreachable by construction.** `usernameRules()` constrains usernames to `/^[a-z0-9._-]+$/` — no `@` — and an email must contain one.

So no precedence rule was invented. Inventing one would be worse than useless: it would imply an ambiguity that doesn't exist, and the next reader would trust it. The docblock states the disjointness and names the invariant instead.

## What I changed after the hand-back, and why it isn't scope creep

The worker enumerated every write path to `users.username` as asked, and **reported the gap plainly instead of glossing it**: `UserFactory` validates nothing, and — the part that matters — `username` had **no model-level enforcement at all**, unlike `email`, whose `saving()` hook folds unconditionally and whose own comment calls itself *"the one hook nothing can bypass"*.

That gap mattered more after this item than before it. `AuthenticateUser` now dispatches on `str_contains($identifier, '@')`. The dispatch is unambiguous **only** while the character sets stay disjoint, and the sole thing keeping them disjoint was every write path remembering to validate.

**The consequence of losing that is not a failed login.** A username equal to another account's email sends the typing user to *that account's* row. That is an account-takeover shape, and `decision/0034` records the precedent: a gap that was harmless only because nothing exercised it must be closed by the item that starts exercising it.

`User::booted()` now folds the username and raises `UsernameWouldBeAmbiguous` on an `@`, beside the email fold and for the same stated reason. Two tests cover it.

## The mutation entry, shaped by last item's lesson

`user-model-refuses-an-at-sign-in-a-username` removes the guard but **leaves the fold**, so the mutated code is valid PHP that persists an `@` username. The test then fails on **data**, not on an exception.

That distinction is exactly what `item/search-filters` had to learn the hard way one item ago, when its FTS5 guard failed on a placeholder desync and would have proved SQLite counts placeholders. Applied here before CI had to teach it again. Simulated locally: mutated file passes `php -l`, guard gone, fold intact.

All **45** `mutations.json` entries re-verified: every `remove` matches its file exactly once.

## Reported, not fixed

- **`UserFactory` still validates nothing.** It no longer endangers the invariant, because the hook catches it — which is the point of putting the guard at the model rather than adding a sixth place that must remember.
- Fortify's lockout messaging wasn't audited for wording that still says "email"; the throttle key is unaffected since the input name stays `email`.

## Mutation to follow

Predicted load-bearing: `runVerify()`'s login filling `ADMIN_USERNAME` into the relabelled field, then `waitForURL(pathname !== '/login')`. **Prediction, not claim** — rule 1 will refuse the Completed record without it.

Closes #75

---


abdullah-life:
## Mutation verdict: negative — and it isolated the username branch, not login

turbophp/doccum#169 pinned `$column` to `'email'` and left everything else — the folding, the model guard, and the blade's `type="text"` field with its "Username or email" label.

```
[setup] logged in with the password doccum:user:reset-password set -- the link genuinely worked
[setup] forgot-password says mail is not configured, byte-identically, for a real account and for no account at all
[verify] logging back in with the USERNAME against the replacement container
##[error]container-smoke (verify) failed: page.waitForURL: Timeout 15000ms exceeded.
```

**The whole setup phase passed, including logins by email.** The reset-password check logs in with a freshly set password and succeeded. Only the verify phase's username login was stranded on `/login`.

So the verdict is sharper than "the smoke notices a broken login": **every email login still worked, so this isolates username resolution specifically.** That is what makes the mutation informative — had I reverted the blade too, the browser would have rejected the input before the server was consulted and the run would have proved much less.

All six `test` jobs went red too, as the mutation PR predicted — `AuthenticationTest` covers login by username directly. Recorded that way rather than presented as the smoke being the sole guard.

## One flaw fixed (`9ff8a30`)

The failure arrived as a bare `page.waitForURL: Timeout 15000ms exceeded`, readable only because the console line above happens to say USERNAME. **Rule 3 says a failure that names no assertion is not evidence**, and `item/home-dashboard` fixed this same shape two items ago.

It now names the account and carries the distinction the mutation revealed — that every earlier email login passed, so this is the username branch of `AuthenticateUser` and not login at large. The assertion is unchanged (same fill, same `waitForURL`, same timeout), so #169's verdict still describes what this code asserts.

#169 is closed unmerged once this round is green.

---

PR #176, opened 2026-09-18
File preview: step through the listing, and say when there is nothing to show

abdullah-life:
Follows #163, which added the preview dialog itself. Three changes, all found by opening the thing rather than by reading it.

## Stepping through the listing

Arrow keys and a pair of header controls move to the adjacent file, with a "2 of 3" position beside them — the only way to know a preview has more behind it.

The order comes from `filesQuery()`, the same query the table renders, so **"next" means the next row you can see**. That differs from the next id the moment anything is sorted by size or owner, and there's a test that sorts by size and requires stepping to follow the screen rather than the database.

Each step goes through `preview()`, so it is authorised exactly as opening one is — a neighbouring row is not evidence of anything. There's a test for that too.

## A file with no stored version says so

Before this, such a file framed `FileDownloadController`, which answers 404 for precisely that case — so the dialog rendered a 404 page inside itself, indistinguishable from the preview being broken.

Found by opening one, not by a test: five rows in local data were in that state.

## Container smoke coverage

This repository's rule is that a UI item is not done until the smoke drives it, and the row actions had none. The new check uploads a file, clicks the **View** icon, and reads the frame's own document to require the uploaded text *inside* it — not merely that a dialog appeared, which an `<iframe>` pointing at a 404 satisfies perfectly well.

It is also the first check to drive an icon action, so it establishes that the aria-labelled icons reach their methods in the built image.

### Its mutation, with the limit stated

```
guard present -> pass: frame contains the file bytes
guard deleted -> fail: locator.waitFor: Timeout 8000ms exceeded
restored      -> pass
```

Measured against the dev server, **not** the container. So it proves the assertion discriminates, and does not yet prove it discriminates in the image. The first CI run of this check establishes the other half, and the docblock says that rather than implying more.

## Checks

- 774 tests / 1649 assertions green
- `pint` clean, `phpstan` 0 new findings
- Verified in a real browser: PNG renders at 697×697, AVIF at 1184×300, text renders in the frame, Escape closes


abdullah-life:
Merged `main` into this branch (`0835c53`) — it had **no CI at all**, not red and not pending, because it was un-mergeable.

That is worth naming rather than just fixing: GitHub cannot build a merge commit for an un-mergeable PR, so the `pull_request` event never fires and no workflow runs. `decision/0047` recorded the same signature the last time it happened. A PR that looks like it is waiting for CI may be waiting for a conflict nobody has noticed, and this one had been sitting since `dbb8374` while `main` moved eleven commits.

Nothing here was rewritten — a merge commit, so your checkout stays valid and you can pull.

## What conflicted, and how

`container-smoke.mjs` and `routes/web.php` **auto-merged cleanly**; both sides only appended. The merged smoke has both sets of checks in the setup phase, one copy of the `clickAndWaitForLivewire()` helper, and `/files/{file}/preview` alongside the new `admin/*` routes.

`.github/mutations.json` conflicted, because both sides appended to the same array. Resolved as a **union**, per `decision/0047` — keeping this branch's `previewed-html-is-served-as-text-not-html` and `main`'s fourteen since `dbb8374`.

All **55** entries were then re-verified against the merged tree: every `remove` string matches its file exactly once. That check is the only thing that makes a union correct, since half the entries now point at code the other side wrote. `node --check` and `php -l` clean on the merged files.

## Two things landed on `main` that touch your diff

- **`clickAndWaitForLivewire()`** (#182 / `decision/0053`) — a bare `waitForLoadState('networkidle')` registered *after* a click can resolve before Livewire has even issued its XHR, so the next `tinker` read wins the race. Your preview check clicks the View icon; if it later reads state straight after a click, it wants this helper.
- **Only `flux:button` is proven to forward an arbitrary attribute in the built image** (`decision/0056`). A `data-test` on a `flux:input` has no precedent for reaching the inner `<input>`; existing smoke checks locate those by `getByLabel()`.

## On your own mutation note

You wrote that the preview mutation was measured against the dev server and not the container, and said so rather than implying more — that is the right call, and this run is the half that was missing. If `image` goes green here, the check has still only been shown to discriminate on the dev server; the container half needs the deliberately-broken image. I am happy to run that mutation once CI reports, if you would rather not spend the cycle.

CI should start within a minute or two now that the merge commit is buildable. I have not reviewed the substance of the diff yet and have not merged anything to `main`.

---

abdullah-life:
CI ran for the first time after the merge, and it found something. **12 of 13 green; `image` red** — [run 35336474153](https://github.com/turbophp/doccum/actions/runs/35336474153/job/105572361932).

I am not pushing a fix: this is a design call that belongs to you, not to me. Diagnosis below, and it is complete — you should not need to re-derive it.

## The failure

`checkTrashRemovesFileFromListingAndSearch` — a **pre-existing** check, unchanged on both sides of the merge (that part of `container-smoke.mjs` auto-merged untouched):

```
[setup] search finds DoccumSmokeTrashTarget.txt before it is trashed
##[error]container-smoke (setup) failed: locator.click: Timeout 30000ms exceeded.
  - waiting for locator('[data-test="trash-file-button"]')
    - locator resolved to <button … wire:click="trashFile" data-test="trash-file-button" …>
  - attempting click action
      - element is visible, enabled and stable
      - <aside data-test="preview-properties" …> from
        <div x-show="true" data-test="preview-modal" class="fixed inset-0 z-50 …"> subtree
        intercepts pointer events
    - retrying click action
    56 × waiting for element to be visible, enabled and stable
```

The button is found, visible, enabled and stable. It is never clickable, because **your preview modal is open on top of it** — a `fixed inset-0 z-50` overlay whose `preview-properties` aside sits over the detail panel and swallows the pointer events. Playwright retried for the full 30s.

The access log immediately before confirms the sequence: `GET /files/1` → a Livewire update → **`GET /files/2/preview`**, then nothing but retries.

## Why

Your `9092794` ("a legible sidebar, **click-to-open**, and properties beside the preview") changed what clicking a row does. That check clicks a row to reach the detail panel and then clicks Trash inside it. With click-to-open, the first click now opens the preview modal instead, and the modal is over the thing the second click wants.

So this is not the merge's doing — it is a real, reproducible consequence of the row-click change, on a check that was already there. What the merge did was let CI run at all, which is why it surfaced now rather than after landing on `main`.

## Two fixes, and I do not think they are equivalent

1. **The check closes the preview first** (Escape, or the modal's own close control) before reaching for the detail panel. Smallest diff. Correct only if opening a preview on a plain row click is the behaviour you intend.
2. **Row click does not open the preview** — the View icon does, and a row click keeps selecting into the detail panel as before. Bigger, but it keeps the two gestures distinct.

Worth weighing: `checkBulkTrashLeavesUnselectedFilesAlone` uses `clickFileRow()` with modifier keys to build a multi-row selection. If a plain row click now opens a modal, selection and preview are competing for the same gesture, and that check is likely to hit the same wall for the same reason once this one is past. That leans me toward (2), but it is your item and your call.

## One more thing the merge brought in

`clickAndWaitForLivewire()` is now in this file (#182 / `decision/0053`). A bare `waitForLoadState('networkidle')` registered *after* a click can resolve before Livewire has even issued its XHR — your preview check clicks the View icon, so if it reads any state straight after a click, it wants that helper rather than a bare settle.

Everything else is green: `guards` (55 entries, all load-bearing), all six `test` jobs, `lint`, `build assets`, `validate ledger`.

---

